PR #1, opened 2023-06-29
Add Danger plugin to check if a PR with new classes also has unit tests

iangmaia:
This PR adds a Danger plugin to check whether newly added classes in a Pull Request also have corresponding unit tests.

It is possible to configure the check with custom parameters to ignore certain classes (`classes_exceptions`), certain subclasses (`subclasses_exceptions`) and a label in the PR (`bypass_label`) that will bypass the checks in the PR.


lib/dangermattic/plugins/unit_test_pr_checker.rb
```diff
@@ -0,0 +1,131 @@
...
+    CLASSES_EXCEPTIONS = [
+      /ViewHolder$/,
+      /Module$/,
+      /ViewController$/
+    ].freeze
+
+    SUBCLASSES_EXCEPTIONS = [
+      /(Fragment|Activity)\b/,
+      /RecyclerView/
+    ].freeze
```
AliSoftware: Those regexes and default values seem strongly tied to Android, while I see that when you ported this rule from Tumblr-Android to here you attempted to make it work for iOS too.

 - Keywords like `abstract` and `protected` don't exist in iOS codebase, het the `NON_PRIVATE_CLASS_DETECTOR` checks for them
 - I _think_ that apart from that, the Regexes will probably work for Swift. But they won't work for ObjC code… yet [you seem to allow `*Tests.m` files](https://github.com/Automattic/dangermattic/pull/1/files#diff-1ad026b0a66eccfe9c8506b7032742c1dfe8e079ebbc6778348402333252d14fR9) to be analyzed still
 - The `CLASSES_EXCEPTIONS` and `SUBCLASSES_EXCEPTIONS` constants used as default values are only making sense for Android

So overall I think we should:
1. Either make that `Plugin` only support Android and rename it `android_unit_test_pr_checker` then plan for a separate Plugin for iOS with its own Regex detectors and default values — even if both end up based on the same internal implementation, but exposing it as two separate plugins calling that common implementation with reasonable defaults and custom detectors…
2. Or make the code for this plugin more flexible to make it easier to have reasonable defaults to support both iOS and Android…
3. Or brainstorm a nicer API that exposes a configuration object — with helper constructors to provide good defaults for iOS and Android

spec/unit_test_pr_checker_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+
+        allow(@my_plugin.github).to receive(:pr_labels).and_return(['my_label'])
+      end
+
+      it 'shows that an Android PR needs tests' do
+        changes_dict = {
+          'File1.java' => 'import java.utils.*;\n\n public class Abc { public static void main(String[] args) { println(''); } }',
+          'project/src/androidTest/java/org/test/ToolTest.kt' => 'class ToolTest { void testMethod() {} }',
+          'File2.kt' => 'class Abcdef(name: String) { public void testMe() { println(''); } }'
+        }
```
AliSoftware: Those look like full file contents, as opposed to `diff` results like I'd expect from `git.diff`. I saw your implementation of `run_in_repo_with_diff` below but that means we'd be unable to test what happens if a class definition existed before the diff and was removed by the PR.

For example, imagine the case where before the PR we had a `class Foo` and tests for that `class Foo`. But the PR deletes the tests–while keeping the `class Foo` around. We want the plugin to detect that there's now no more tests for `Foo` and warn about it.

Or imagine that the PR removes both `class Foo` and its tests, we would want to make sure that the plugin see those as deletions, as opposed to accidentally (due to a flaw in our implementation, which is what our Unit Tests are checking for) see any changes as additions, even line removals.

Or finally, imagine if the definition of `class Foo` already existed before the PR, but the PR just moved that class definition into a new file (e.g. refactoring, or moving from a `.java` legacy code to `.kt` code…).

Those are all examples we need to test for, but your `run_in_repo_with_diff` helper would not be able to cover (and your current unit tests don't)

spec/unit_test_pr_checker_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+          `git commit -m "add files"`
+
+          # adds changes to the previously commited files to create a diff
+          changes_dict.each do |key, value|
+            File.open("#{dir}/#{key}", 'w') { |f| f.write(value) }
+          end
+
+          g = Git.open('.')
+          yield g
+        end
```
AliSoftware: While this should work, I'm unsure if it's a good thing to create a real git repo in a temp dir and do the real git actions (`git add`, `git commit`), as opposed to create a mock or test_double of the `Git` object that would return the mocked content.

lib/dangermattic/plugins/unit_test_pr_checker.rb
mokagio: > Either make that Plugin only support Android and rename it android_unit_test_pr_checker then plan for a separate Plugin for iOS with its own Regex detectors and default values — even if both end up based on the same internal implementation, but exposing it as two separate plugins calling that common implementation with reasonable defaults and custom detectors…

My two cents: I think we'd be better off exposing dedicated plugins with platform-specific APIs that users can call. Let consumers easily configure the tool. Do all the "hard work" in the tool's backend

spec/unit_test_pr_checker_spec.rb
iangmaia: > Those are all examples we need to test for, but your run_in_repo_with_diff helper would not be able to cover (and your current unit tests don't)

Agree -- I'm gonna refactor `run_in_repo_with_diff` and the tests to prefer mocking over the "real" git diffs to also help covering these additional cases.

spec/unit_test_pr_checker_spec.rb
iangmaia: This was something I remember seeing in `release-toolkit` and thought would be cleaner/more readable than mocking diff strings; however, the test becomes a sort of integration test as well. I agree, I'll refactor it to use mocks and the tests as well, also to cover the other cases you mentioned in the other comment 👍

lib/dangermattic/plugins/unit_test_pr_checker.rb
iangmaia: > Those regexes and default values seem strongly tied to Android, while I see that when you ported this rule from Tumblr-Android to here you attempted to make it work for iOS too.

Indeed a "one size fits all" solution probably can get hard to follow here, as it will require convoluted regexes and constants. I thought it could be simple enough for this but it also makes it harder to maintain and extend.
I'm agreeing with @mokagio and thinking that separate plugins (with some shared code) will be the simplest in this case.

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
```diff
@@ -0,0 +1,131 @@
...
+            end
+          end
+        end
+      end
+
+      # We only want newly added classes, not if class signature was modified or line was moved
+      violations.reject! { |v| removed_classes.include?(v.classname) }
+      # For each remaining candidate, only keep the ones _not_ used in a new test
+
+      violations.select { |v| added_test_lines.none? { |line| line =~ /\b#{v.classname}(Tests?)?\b/ } }
```
iangmaia, line 114: I slightly changed the regex to cover when a test class name ends with `Test` or `Tests`.

spec/unit_test_pr_checker_spec.rb
iangmaia: > For example, imagine the case where before the PR we had a class Foo and tests for that class Foo. But the PR deletes the tests–while keeping the class Foo around. We want the plugin to detect that there's now no more tests for Foo and warn about it.

Since this plugin is checking only the PR diff, I believe this case won't be covered by the nature of Danger?
Otherwise, I changed the tests a bit to cover other scenarios and added more test data.

But I was thinking we could eventually sophisticate this check with another plugin by using build artefacts and checking whether the test coverage has reduced with a new PR (which of course implies on saving the coverage state somewhere), having a similar effect as just checking the PR diff.

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
```diff
@@ -0,0 +1,128 @@
+# frozen_string_literal: true
+
+require_relative 'utils/git_utils'
+
+module Danger
+  # Plugin to detect classes without Unit Tests in a PR.
+  class AndroidUnitTestPRChecker < Plugin
```
AliSoftware: [Open Question] 

I'm started to ponder if we should postfix all our plugins with `…PRChecker`. After all, the main (only?) way we plan to use `dangermattic` and its _danger_ rules is to check PRs. And getting rid of that suffix will make some room for more explicit rule names — especially since the name of those plugins are what the client repo ends up using in their `Dangerfile` (e.g. `android_unit_test_pr_checker.check_missing_tests` 🤔

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
```diff
@@ -0,0 +1,128 @@
...
+      /ViewHolder$/,
+      /Module$/
+    ].freeze
+
+    SUBCLASSES_EXCEPTIONS = [
+      /(Fragment|Activity)\b/,
+      /RecyclerView/
+    ].freeze
+
+    UNIT_TESTS_BYPASS_PR_LABEL = 'unit-tests-exemption'
```
AliSoftware: ~I think this should be a parameter, because each client repo that will use this Danger plugin will likely use different GitHub labels (and to my knowledge, only Tumblr iOS & Android have that `unit-tests-exemption` label set up).~

~Overall I'd argue for all our repos to have the same GitHub Labels when it comes to labels that interact with our tooling like Danger, for consistency across repos. But I'm not sure if we want to hardcode the label name to the point that it can't be changed for repos that don't follow that convention.~

I just realized after writing all that, that the `bypass_label` is in fact already a parameter to the `check_missing_tests` public method and plugin API 👍  Maybe we should rename that constant to `DEFAULT_BYPASS_PR_LABEL` instead tho, to make it clearer? (And then probably same for `CLASSES_EXCEPTIONS` and `SUBCLASSES_EXCEPTIONS`)

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
```diff
@@ -0,0 +1,128 @@
...
+    # @example Check missing unit tests
+    #   check_missing_tests()
+    #
+    # @example Check missing unit tests excluding certain classes and subclasses
+    #   check_missing_tests(classes_exceptions: [/ViewHolder$/], subclasses_exceptions: [/RecyclerView/])
+    #
+    # @example Check missing unit tests with a custom bypass label
+    #   check_missing_tests(bypass_label: 'BypassTestCheck')
+    def check_missing_tests(classes_exceptions: CLASSES_EXCEPTIONS, subclasses_exceptions: SUBCLASSES_EXCEPTIONS,
+                            bypass_label: UNIT_TESTS_BYPASS_PR_LABEL)
```
AliSoftware: This had me wonder, how do we plan to document all our various plugins exposed via `dangermattic`? 🤔 💭 

For example, say that I'm a client repo (e.g. WordPress-Android) interested to know what Danger plugins I can use from `dangermattic`, but also and more importantly how to call them (what is the name of the plugins and of their public API methods that I can call from my `Dangerfile`), will my only option be to have to look at the source code to find out?

 - One option could be to write a `README.md` in each `lib/dangermattic/plugins/` folder (and then maybe reference those in the main `README.md` at the repo's root) that would all have a similar format/structure telling what the plugin does, when and how to use it, and examples.
 - Another option would be to make sure that the YARD doc for this gem gets properly generated, and make it easy to access it (direct link in the repo's root `README.md`?).

The second option seems appealing as that will allow us to have the documentation nearest to the code and make it easier to maintain, but having a `README` with examples and stuff would also be nice. So maybe we should go with a mix of both?

💡  We could maybe take inspiration from how other _danger_ plugins are documented using both YARD and a README with examples. See https://www.rubydoc.info/gems/danger-changelog for example.

Anyway, all of this is probably for a separate PR (where we'd maybe add a CI step to validate the YARD doc is complete, and without missing documentation for public methods of `Plugin` subclasses at least, and to warn us about doc potentially being out of date with the method signature — like parameters referenced in the doc and method not matching anymore — etc, and to come up with a structure for our `README` files…), but I figured this was worth starting the discussion at some point 🙂

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
```diff
@@ -0,0 +1,128 @@
...
+      end
+    end
+
+    private
+
+    ClassViolation = Struct.new(:classname, :file)
+
+    # @param [Git::Diff] git_diff the object
+    # @return [Array<ClassViolation>] An array of `ClassViolation` objects for each added class that is missing a test
+    def find_classes_missing_tests(git_diff:, classes_exceptions:, subclasses_exceptions:)
```
AliSoftware: I'm guessing that in the future (out of scope of this PR), we might want to migrate that helper method into `utils/` and make it a bit more generic (not referencing global constants like `NON_PRIVATE_CLASS_DETECTOR` and `ANY_CLASS_DETECTOR`) so that it can be mutualized and reused for a future `ios_unit_test_pr_checker.rb` plugin too 🤔  But we can keep that DRY-ing for later, especially for the PR that will introduce the iOS plugin counterpart of this one, I guess 🙂

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
```diff
@@ -0,0 +1,128 @@
...
+            end
+          end
+        end
+      end
+
+      # We only want newly added classes, not if class signature was modified or line was moved
+      violations.reject! { |v| removed_classes.include?(v.classname) }
+      # For each remaining candidate, only keep the ones _not_ used in a new test
+
+      violations.select { |v| added_test_lines.none? { |line| line =~ /\b#{v.classname}(Tests?)?\b/ } }
```
AliSoftware: I don't think this is correct. This code is supposed to check, amongst all the lines added to test files, if the word (`\b` = word boundary) corresponding `v.classname` appears anywhere in those lines of code from Tests.

So for example, if, amongst the lines added in test files, you have `new MyClass(foo)`, then we consider that the class `MyClass` was used somewhere in an added test line, so that means that yes, the PR added a new class definition/implementation called `MyClass`, but that `MyClass` was also used in one of the unit tests, so that one is good and not a violation we should keep.

 - This has nothing to do with the _file_ names being used (because those classes could be used in the tests by adding new tests in an _existing_ `SomeFeatureTest.kt` file anyway, as opposed to always having to be used in a file that has to be named `MyClassTest.kt`)
 - We don't want a violation to be excluded if the PR introduces a new class named `Foo` inside `Foo.kt`, and then introduces a new file `FeatureTests.kt` whose content includes `class FooTests {}` but without any actual test calling/referencing the `Foo` class itself (like `new Foo()`). But that's what you modified regex would allow though.

```suggestion

      # For each remaining candidate, only keep the ones _not_ used in a new test
      violations.select { |v| added_test_lines.none? { |line| line =~ /\b#{v.classname}\b/ } }
```

spec/android_unit_test_pr_checker_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+
+        @plugin.check_missing_tests
+
+        expect(@dangerfile.status_report[:errors].count).to eq 5
+      end
+
+      it 'does not show errors when new classes have corresponding tests' do
+        changes_dict = {
+          'Abc.java' => 'import java.utils.*;\n\n public class Abc { public static void main(String[] args) { System.out.println(""); } }',
+          'project/src/androidTest/java/org/test/AbcTests.java' => 'class AbcTests { public void abcTest() { sut.main([]) } }',
```
AliSoftware: I don't think this should be considered a test for the class `Abc`?
While the test class is called `class AbcTests`, none of the code in that test file references the `Abc` class (like `Abc.main([])`.

That is, except I'm missing some Java magic that I'm not aware of (which is possible, as I never wrote any serious Java personally…) which would make the `sut` variable magically be defined by the Java compiler on test files as being the class with the same name as the test class `AbcTests` but without the `Tests` suffix…? Which is what would make that Java code compile and be considered a test on the `Abc` class itself?

spec/android_unit_test_pr_checker_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+      it 'show errors when we remove test classes for classes we refactored' do
+        added_classes = {
+          'Abc.kt' => 'import java.utils.*;\n\n public class Abc { public static void main(String[] args) { System.out.println(""); } }',
+          'Polygon.kt' => 'data class Polygon(sides: Int) { fun draw() {} }',
+          'TestsINeedThem.kt' => 'public open class TestsINeedThem { fun testMe2() { } }'
+        }
+
+        removed_tests = {
+          'project/src/androidTest/java/org/test/AbcTests.java' => 'class AbcTests { public void abcTest() { sut.main([]) } }',
+          'project/src/androidTest/java/org/test/PolygonTest.java' => 'class PolygonTest { void drawTest() {}',
```
AliSoftware: Same question, how is this supposed to be considered a unit test on the `Polygon` class when none of the code in this Unit Test and test file references the `Polygon` class itself? The sole fact that the `PolygonTest` class is defined in a `*Test.java` file and called `PolygonTest` isn't enough imho to make us consider that this is testing the `Polygon` class just because of sheer naming conventions?

So even if we did not remove that file (if the PR didn't have those as part of `removed_tests` / deleted files diff), and we still had this `PolygonTest.java` file not deleted, we should already have a violation for `data class Polygon` not having any Unit Tests… no?

spec/android_unit_test_pr_checker_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+          'project/src/androidTest/java/org/test/TestsINeedThem.java' => 'class TestsINeedThem { public void testMe() {} }'
+        }
+
+        diff = generate_add_diff(added_classes) + generate_deleted_diff(removed_tests)
+        allow(@dangerfile.git).to receive(:diff).and_return(diff)
+
+        @plugin.check_missing_tests
+
+        expect(@dangerfile.status_report[:errors].count).to eq 3
```
AliSoftware: I'd suggest your Unit Tests to not just verify if the number of errors is the expected one, but to also check if the error messages that are reported are the expected one.

Because if we change the implementation one day in some breaking way, but by sheer change that makes one violation that was triggered before not be triggered anymore… while it also makes another violation that was not triggered before now be triggered… you'd end up with a `n+1-1` situation that would make your errors count unchanged and would have your spec still pass, making you think it still works while your refactoring of the implementation in fact broke things.

spec/android_unit_test_pr_checker_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+          --- a/#{file_path}
+          +++ /dev/null
+          @@ -1 +0,0 @@
+          -#{content}
+          \\ No newline at end of file
+        PATCH
+
+        OpenStruct.new(type: 'deleted', path: file_path, patch: diff_str)
+      end
+    end
```
AliSoftware: That just made me think: we probably want to also test the case of a diff that contains a mix of both added and removed lines inside the same file.

Typical example is for checking that if a class definition, or a test using that class, moved within a file, we will not incorrectly consider it an addition or a deletion, but correctly [`reject!` it](https://github.com/Automattic/dangermattic/pull/1/files#diff-7006d15ff302c846b3271adb5945d3fd455d7c13d35fca97b46444e9f3ede9e4R108) from the violations instead.

e.g. say we just moved the `class Foo` implementation some place later in the `Foo.kt` file…
```diff
--- a/Foo.kt
+++ b/Foo.kt
@@ -3 +0,0 @@
- class Foo {
-   fun demo() { }
- }
@@ -27,0 +3 @@
+ class Foo {
+   fun demo() { }
+ }
```

AliSoftware:
PS: Could be worth rebasing this PR on top of `trunk` now that #6 has landed, so that the CI would run the specs you added in this PR and validate they are all green.

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
iangmaia: This was something I had in mind when I started with the prototype, to have some convention around PR checks to clearly separate them from Post Build checks that will use build artifacts such as coverage report, built files, test results, etc.
Now a bit deeper into the project, I don't think it makes a lot of sense. Hopefully the name of the plugin on its own will be enough to differentiate them; also, most of them will likely be PR checks, as you mentioned, and hopefully their names will be descriptive enough so that users of the Gem will be able to use a plugin whenever appropriate for them.

I can rename the plugin to something like `AndroidUnitTestChecker` or `AndroidMissingUnitTestChecker` or something like `AndroidNeedsUnitTests`. Wdyt?

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
iangmaia: >  Maybe we should rename that constant to DEFAULT_BYPASS_PR_LABEL instead

Agree! Will rename all constants.

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
iangmaia: > 💡 We could maybe take inspiration from how other danger plugins are documented using both YARD and a README with examples. See https://www.rubydoc.info/gems/danger-changelog for example.

Very good point, and I like the idea 👍 

> add a CI step to validate the YARD doc is complete, and without missing documentation for public methods of Plugin subclasses at least, and to warn us about doc potentially being out of date with the method signature

Also nice idea. I think once we get the first few plugins ready on `trunk` we'll have enough material (but still a not very large amount to do this at once) to start a documentation structure and add these validations 👍  I'll add that as a project task, thanks!

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
iangmaia: Indeed, this will be part of the code shared between the iOS and the Android plugins 👍 and I guess we'll also be able to reuse the test scenarios as test examples as well.

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
iangmaia: Ah! I misunderstood the intention here and then that is the reason for a few other points you mentioned below. I thought it was looking for actual test class counterparts of the original classes (which would make the sole word `Classname` a bit odd, but I thought that was the intention if for some reason we had a test class inside a test folder with the exact same name, but it could be a project convention or something like that), not uses of that class.

> We don't want a violation to be excluded if the PR introduces a new class named Foo inside Foo.kt, and then introduces a new file FeatureTests.kt whose content includes class FooTests {} but without any actual test calling/referencing the Foo class itself (like new Foo())

Clear!

spec/android_unit_test_pr_checker_spec.rb
iangmaia: That comes from the misunderstanding you pointed out in [this suggestion](https://github.com/Automattic/dangermattic/pull/1#discussion_r1260934060).

I'll change these tests that have no usage at all (no magic in Java to auto-create a `sut`, unfortunately 🪄😅) to reflect that.

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
AliSoftware: Maybe we could add some `# code comment` above that `\b#{v.classname\b` regex to clarify this for future readers of the code and avoid future confusion?

spec/android_unit_test_pr_checker_spec.rb
iangmaia: That comes from the misunderstanding you pointed out in [this suggestion](https://github.com/Automattic/dangermattic/pull/1#discussion_r1260934060).
I thought the initial plugin was based on the test naming and not usage, but the comment above clarified everything and it all makes sense now 👍

spec/android_unit_test_pr_checker_spec.rb
iangmaia: Good idea! Just pushed some changes and a new test adding a similar diff 👍

lib/dangermattic/plugins/android_unit_test_pr_checker.rb
iangmaia: Rename the plugin to simply `AndroidUnitTestChecker`.

spec/android_unit_test_checker_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+      before do
+        @dangerfile = testing_dangerfile
+        @plugin = @dangerfile.android_unit_test_checker
+
+        allow(@plugin.github).to receive(:pr_labels).and_return(['my_label'])
+
+        stub_const('GitDiffStruct', Struct.new(:type, :path, :patch))
+      end
+
+      it 'shows the right number of errors when a PR adding new classes that do not have corresponding tests' do
```
AliSoftware: With your latest changes, this doesn't just check the _number_ of errors anymore, but the actual errors now
```suggestion
      it 'shows the right errors when a PR adds new classes that do not have corresponding tests' do
```

spec/android_unit_test_checker_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+
+        diff = generate_add_diff(changes_dict)
+        allow(@dangerfile.git).to receive(:diff).and_return(diff)
+
+        @plugin.check_missing_tests(classes_exceptions: classes_to_ignore, subclasses_exceptions: subclasses_to_ignore)
+
+        expect(@dangerfile.status_report[:errors]).to be_empty
+      end
+
+      it 'does nothing when a PR moves code around and' do
```
AliSoftware: moves code around and…?

spec/android_unit_test_checker_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+      end
+    end
+
+    def class_names_match_report?(class_names:, error_report:)
+      error_report.all? do |str|
+        class_names.any? do |class_name|
+          str.include?("Please add tests for class `#{class_name}`")
+        end
+      end
+    end
```
AliSoftware: I might be wrong, but I think this only checks if all errors that are reported contain the name of one of the `class_names`. But that means that the `class_names` array can have a lot more items in it and this will still be true.

```ruby
[1] pry(main)>     def class_names_match_report?(class_names:, error_report:)
      error_report.all? do |str|
        class_names.any? do |class_name|
          str.include?("Please add tests for class `#{class_name}`")
        end
      end
[1] pry(main)> def class_names_match_report?(class_names:, error_report:)
  error_report.all? do |str|
    class_names.any? do |class_name|
      str.include?("Please add tests for class `#{class_name}`")
    end
  end
=> :class_names_match_report?  
[2] pry(main)> error_report = %w[A B C].map { |c| "Please add tests for class `#{c}`" }
=> ["Please add tests for class `A`", "Please add tests for class `B`", "Please add tests for class `C`"]
[3] pry(main)> class_names = %w[A B C D E F]
=> ["A", "B", "C", "D", "E", "F"]
[4] pry(main)> class_names_match_report?(class_names: class_names, error_report: error_report)
=> true
```

Instead we need to test both directions, i.e. that not only all errors reported in `error_report` mention one of the expected `class_names`, but also that all the `class_names` are reported in `error_report`. And even further, ideally we want the list to match the counts exactly — i.e. that if `error_report` contains 4 errors but 2 of those errors are both for class `B`, then maybe we should pass `class_names: %w[A B B C]` to explicitly indicate we expect B twice.

---

Finally, personal preference but I think I'd find it easier to read at call site if that method included the `expect` call directly — as opposed to the method returning a boolean and then the call site having to do `expect(result).to be(true)` 🤷 

Maybe something like this would work then:
```ruby
expect(class_names.length).to eq(error_report.length)
class_names.zip(error_report).each do |cls, error|
  expect(error).to include "Please add tests for class `#{cls}`"
end
```

iangmaia: > I might be wrong, but I think this only checks if all errors that are reported contain the name of one of the class_names. But that means that the class_names array can have a lot more items in it and this will still be true.

> Finally, personal preference but I think I'd find it easier to read at call site if that method included the expect call directly

On this case, as we have full control over the mocked data and the test itself must be correct, I don't think this would be an issue (e.g. the person writing the test knows that the classes are the classes that should be in the diff).
But the usage of `zip` is super neat and would indeed account for errors on setting up the data, so I think it is nicer in the end 👍  I also agree with using the expect in the helper function, as long as it also makes it clear that an expectation will be used there (I'll name it accordingly).

Thanks for the suggestion! 👍

AliSoftware: > On this case, as we have full control over the mocked data and the test itself must be correct, I don't think this would be an issue (e.g. the person writing the test knows that the classes are the classes that should be in the diff).

The main problem is if we update the implementation in the future and accidentally break things, leading to `error_report` only reporting errors for `A` and `B` but stopping to report errors for class `C`. The unit tests that was passing `class_names: %w[A B C]` would then still pass (because `error_reports.all? { … class_names.any? … {}}` would still pass with all of `A` and `B` passing the `any?` test) and thus your test won't do its job of detecting the regression.

iangmaia: Ah that's true, indeed it wouldn't catch a regression like that 👍  I've already updated the code.

AliSoftware (approved):
I'd love @ParaskP7, as our Android guru, to suggest additional, tricker Kotlin (and Java?) syntaxes to use in our unit tests, both for declaring classes and for using them in examples.

I'm especially thinking about declarations that include annotations, constructor parameters, inheritance clauses, and a mix of all that in the `class Foo…` line declaring the class, so that we have more convoluted examples to test against and validate that our RegEx detecting class declarations account for those tricky cases too.

ParaskP7:
👋 @AliSoftware , thanks for the ping, plus awesome job so far @iangmaia ! ❤️ 

> I'd love @ParaskP7, as our Android guru...

😆 Not at all, but thanks! 😊 

> ...to suggest additional, tricker Kotlin (and Java?) syntaxes to use in our unit tests, both for declaring classes and for using them in examples.

Thinking, first, let me clone and have a quick look at `android_unit_test_checker.rb` and its corresponding spec class. 🥇 
 🤔

> I'm especially thinking about declarations that include annotations, constructor parameters, inheritance clauses, and a mix of all that in the class Foo… line declaring the class, so that we have more convoluted examples to test against and validate that our RegEx detecting class declarations account for those tricky cases too.

From what I understood, let me provide you below with a list of example classes from [WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) for you to check against. You could most probably add them to your exceptions list as those can't be unit tested, at least not in any straightforward way, here goes:

- Any [Compose](https://developer.android.com/jetpack/compose) related files:
    - How: You could search for `import androidx.compose.*` imports, or the `@Composable` annotation on top of any function.
    - Examples: [PrimaryButton.kt](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/ui/compose/components/buttons/PrimaryButton.kt#L30)
- Any [Custom View](https://developer.android.com/develop/ui/views/layout/custom-views/custom-components) related files:
    - How: You could search for the `*View.[kt/java]` file suffix.
    - Examples: [BadgedImageView](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/widgets/BadgedImageView.kt#L25) + [DiffView](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/widgets/DiffView.kt#L17) + [NestedWebView](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/widgets/NestedWebView.kt#L18)
- Any [Service](https://developer.android.com/guide/components/services) related files:
    - How: You could search for the `*Service.[kt/java]` file suffix:
    - Examples: [GCMMessageService](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/push/GCMMessageService.java#L29) + [InstallationReferrerService](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/analytics/service/InstallationReferrerService.java#L22)
- Any [Content Provider](https://developer.android.com/guide/topics/providers/content-provider-basics) related files:
    - How: You could search for the `*Provider.[kt/java]` file suffix:
    - Examples: [QueryContentProvider](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/provider/query/QueryContentProvider.kt#L7)
- Any [Broadcast Receiver](https://developer.android.com/guide/components/broadcasts) related files:
    - How: You could search for the `*Receiver.[kt/java]` file suffix:
    - Examples: [InstallationReferrerReceiver](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/analytics/receiver/InstallationReferrerReceiver.java#L10)
- Any [Extension](https://kotlinlang.org/docs/extensions.html) related files: 
    - How: You could search for the `*Extensions.kt` file suffix, `*Ext.kt`, or even without the plural `s`, something like the `*Extension.kt` file suffix.
    - Examples: [ViewExtensions.kt](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/extensions/ViewExtensions.kt#L4) + [CrashLoggingExt.kt](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/crashlogging/CrashLoggingExt.kt#L4)
- Any `Helper` related files:
    - How: You could search for the `*Helpers.[kt/java]` file suffix, or even without the plural `s`, something like the `*Helper.[kt/java]` file suffix.
    - Examples: [CoroutinesHelper.kt](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/coroutines/CoroutineHelpers.kt#L4) + [SupportHelper.kt](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/support/SupportHelper.kt#L17)
- Any `Feature Flag` related files ([WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) specific, but maybe applies to other such patterns on various repos):
    - How: You could search for the `*FeatureConfig.[kt/java]` file suffix.
    - Examples: [BloggingPromptsFeatureConfig](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/config/BloggingPromptsFeatureConfig.kt#L5)
- Any `Wrapper` related files:
    - How: You could search for the `*Wrapper.[kt/java]` file suffix.
    - Examples: [AnalyticsUtilsWrapper](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/analytics/AnalyticsUtilsWrapper.kt#L23)
- Any `Table` related files:
    - How: You could search for the `*Table.[kt/java]` file suffix.
    - Examples: [BlockedAuthorTable](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/datasets/BlockedAuthorTable.java#L9)
- Any `Enum` related files:
    - How: You could search standalone `enum` classes.
    - Examples: [BlavatarShape](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/image/BlavatarShape.kt#L4)

Is the above what you are looking for, does that help? 🤞 

-----

This is just a quick assemble of things I would consider for now. I am very sure I might be missing some, but maybe those are not that crucial, I guess we could keep modifying/adding to our exceptions list going forward, right? 🤔

AliSoftware:
> This is just a quick assemble of things I would consider for now. I am very sure I might be missing some, but maybe those are not that crucial, I guess we could keep modifying/adding to our exceptions list going forward, right? 🤔

Thanks a lot, these are very useful inputs and examples of things we could consider adding exceptions for!
In the examples you linked, I notice two main categories:
1. Exceptions we can make based on the class name suffix (`*Components`, `*Service`) or the parent class they inherit
2. Exceptions based on other parts of the declaration, like if the declaration has a specific annotation (e.g. `@Composable`), or is an `enum class` and not just a `class`.

For case (1), we already kinda have some logic around it (with our `DEFAULT_SUBCLASS_EXCEPTIONS` and the `subclass_exceptions` parameter that we can provide when calling this Danger check)… but in its current form, our logic does not support multi-line statements (because our RegEx parses the code line-by-line), so this won't work for cases where the inheritance clause is not on the same line as the class name.

For case (2), We don't have any logic in our current implementation to account for that (both because our RegEx doesn't capture annotations when trying to match the line, but also once again because our RegEx is single-line-constrainted)

That being said, it feels like a lost cause to re-implement kind of a Kotlin code parser with Regular Expressions alone, especially if we want to support all possible syntaxes for a class declaration with all its possible annotations, `constructor(…)` clauses and possibly more (that's why I needed all those examples, to get a sense of how many variations of the syntax for declaring a class in Kotlin there could be and we'd need to support, so thanks!).
I mean, we could try, but the RegEx might end up quite complex if it needs to support all possible variations + multiline… so maybe another approach (like a char-by-char text scanner / state machine, or a Parser Combinator approach) would be a better fit than a convoluted RegExp?

_@iangmaia TDD and Red/Green/Refactoring would help a lot here, by first using all the examples Petros provided as fixtures for our unit test, have the test fail to detect those and go red, and then find the best way to make the tests go green, thus proving that the new implementation would properly cover all those fixture cases_

---

@ParaskP7 What I was also interested in getting your input on is syntax for class declaration that are not as simple as `class Name : ParentClass` (which should be easy enough for our RegEx to correctly match and extract the class name without risking being tricked), but things more like:

```
@Reusable class AnalyticsUtilsWrapper @Inject constructor(private val appContext: Context)
```
Which includes annotations before and after the `class ClsName`, but also some additional keywords on the declaration line like `constructor` and additional parentheses that also include a type name (`Context`)… all of the things that might be able to trick our RegEx trying to detect that this is a line declaring a new class (as opposed to not recognizing these lines as being class-declaring expressions because of them being more complex than the syntax expected by our limited RegEx), and that this declared class' name is `AnalyticsUtilsWrapper` (and not `Context` or some other word of this line that it would accidentally pick up)

The other one you quoted with feature flags can also make a good example of a tricky fixture to use in our Unit Tests of this plugin to ensure that, despite the tricky syntax, the plugin is still able to detect it as a class and extra the proper class name so that the rest of the plugin logic works as expected:
```
@Feature(BLOGGING_PROMPTS_REMOTE_FIELD, true) class BloggingPromptsFeatureConfig @Inject constructor(appConfig: AppConfig) : FeatureConfig(appConfig, BuildConfig.BLOGGING_PROMPTS, BLOGGING_PROMPTS_REMOTE_FIELD)
```
(Here I intentionally collapsed the whole syntax into a single line, even if I understand that most coding conventions would make this be declared in multiple lines in practice, but my goal here was just to generate an example that would trick the RegEx which operates on a line-by-line scanning)

AliSoftware:
For reference, I just found [this Kotlin spec](https://kotlinlang.org/spec/declarations.html#classifier-declaration), which shows how the syntax for a class declaration can be composed of many optional parts. So maybe there's a a ruby library that would be able to take a [BNF Grammar definition](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) or some grammar definition like the one used in this Kotlin spec, and turn that into a Parser for us, that would account for all the possible combinations allowed by the syntax and be able to extract each component of the declaration in a nice structure for us to use for our logic. i.e such Parser would be able to scan the code, and extract the possible `modifier` in front of `class`, as well as the `primaryConstructor` part, the `deletationSpecifiers`, etc… from the class declaration, present that in a nice Ruby object, and then we'd be able to write easier-to-understand logic in our rule like `if modifiers.classModifiers.include?(:enum)` or `if  modifiers.annotations.include?('Component')`?

That will obviously make the implementation of that Plugin quite more complicated than just using a RegExp. But that would also avoid a lot of false positives from the plugin/rule, especially with multi-line declarations that might declare the inheritance from `View` and similar in a different line that would be missed by our current RegExp otherwise…

(Again, TDD will be your friend here, to first describe your expectations in tests, then experiment with implementations until you find a way to make all of them go green)

ParaskP7:
👋 @AliSoftware !

> Thanks a lot, these are very useful inputs and examples of things we could consider adding exceptions for!

💪 

> For case (1), we already kinda have some logic around it (with our DEFAULT_SUBCLASS_EXCEPTIONS and the subclass_exceptions parameter that we can provide when calling this Danger check)… but in its current form, our logic does not support multi-line statements (because our RegEx parses the code line-by-line), so this won't work for cases where the inheritance clause is not on the same line as the class name.

👍 

> For case (2), We don't have any logic in our current implementation to account for that (both because our RegEx doesn't capture annotations when trying to match the line, but also once again because our RegEx is single-line-constrainted)

👍

> I mean, we could try, but the RegEx might end up quite complex if it needs to support all possible variations + multiline… so maybe another approach (like a char-by-char text scanner / state machine, or a Parser Combinator approach) would be a better fit than a convoluted RegExp?

I think I agree.

> @ParaskP7 What I was also interested in getting your input on is syntax for class declaration that are not as simple as class Name : ParentClass (which should be easy enough for our RegEx to correctly match and extract the class name without risking being tricked), but things more like:
> ...
> The other one you quoted with feature flags can also make a good example of a tricky fixture to use in our Unit Tests of this plugin to ensure that, despite the tricky syntax, the plugin is still able to detect it as a class and extra the proper class name so that the rest of the plugin logic works as expected:

Exactly, I think [BloggingPromptsFeatureConfig](https://github.com/wordpress-mobile/WordPress-Android/blob/44282b137085ec230771fc2c87ad9a44610fbeb1/WordPress/src/main/java/org/wordpress/android/util/config/BloggingPromptsFeatureConfig.kt#L5) is a good example of what you are describing above.

But, to be honest, I didn't quite get if you want more examples like that, do you? 🤔

> For reference, I just found [this Kotlin spec](https://kotlinlang.org/spec/declarations.html#classifier-declaration), which shows how the syntax for a class declaration can be composed of many optional parts. So maybe there's a a ruby library that would be able to take a [BNF Grammar definition](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) or some grammar definition like the one used in this Kotlin spec, and turn that into a Parser for us, that would account for all the possible combinations allowed by the syntax and be able to extract each component of the declaration in a nice structure for us to use for our logic.

This is a good reference, but I am afraid that creating such a parser for the purpose of what we are trying to do here might be a bit too much, not sure... 🤷 

> That will obviously make the implementation of that Plugin quite more complicated than just using a RegExp.

This. ⬆️ 

> But that would also avoid a lot of false positives from the plugin/rule, especially with multi-line declarations that might declare the inheritance from View and similar in a different line that would be missed by our current RegExp otherwise…

Maybe, at least for this initial phase of `Dangermattic`, we could deal with a couple of false positives, here and there, see how many we get, gather some data around those, and only then to try and work around it, not sure... 🤷

AliSoftware:
Yeah maybe just keeping the approach of using our current RegEx (maybe will need to adjust some bits to account for some of the examples you provided though) and just making it support multiline (`/…/m` flag, iinm?) would be enough for the first iteration

iangmaia:
Thanks for your suggestions @ParaskP7! Very helpful specially for us to get a bit of "reality check" on what we can expect.
We can use several of your examples for exceptions.

For @AliSoftware: in addition to an exception list based on name / base class, what about paths (implying in packages) to exclude as well? For example, a host project could just ignore `src/main/java/org/wordpress/android/ui/**` instead of having to specify many view naming and base class exceptions.

> That being said, it feels like a lost cause to re-implement kind of a Kotlin code parser with Regular Expressions alone, especially if we want to support all possible syntaxes for a class declaration with all its possible annotations, constructor(…) clauses and possibly more

💯 (and there's also java/kt generics affecting class declarations, data classes, sealed classes...)

That's also why I was also thinking that an alternative (or complementary) plugin could do a similar check but on a higher level, focusing on the code coverage, preventing the code coverage to drop on a new PR compared to base.

> So maybe there's a a ruby library that would be able to take a BNF Grammar definition [...] and turn that into a Parser for us
> [...]
> That will obviously make the implementation of that Plugin quite more complicated than just using a RegExp. But that would also avoid a lot of false positives from the plugin/rule, especially with multi-line declarations that might declare the inheritance from `View` and similar in a different line that would be missed by our current RegExp otherwise…

When I first read it I was wondering if this wouldn't be an overkill, as these libraries tend to be heavy (I just seen that [Tree-sitter has Ruby bindings](https://github.com/calicoday/ruby-tree-sitter-ffi)). But indeed it would make the plugin a lot more robust.
Like we all know, it's easy to start ignoring warnings if they are "always there" or just unreliable -- at the same time, we don't know exactly how big this problem is: we know that we'll miss multi-line declarations, but if that's the main issue we'll encounter, maybe it's enough to improve the RegEx without requiring a reimplementation?

Based on @ParaskP7's comment and this thread, I think there are some things we could do on this plugin going forward:
1. Use @ParaskP7 suggestions and more real examples as test fixtures, possibly refactoring the way the data is being provided currently in the tests (which are just simple strings in dictionaries), or just add some more simple declarations without a refactor. TDD all the way. We can then based on the results here to decide to go for a more complex implementation or not.
2. Add path based exceptions, if we think that makes sense.
3. Replace the current RegEx based implementation with a more sophisticated parser approach to catch e.g. multi-line declarations, generics, annotations, etc.

Since I also need to focus on migrating [what we have in Peril](https://github.com/Automattic/peril-settings) to Danger, I'm thinking if we can keep the initial implementation we have right now and merge it (or just keep the PR open), and revisit this plugin on a 2nd iteration (but before we add an iOS counterpart, as the improvements discussed above will also benefit iOS).

AliSoftware:
Yeah I think that (1) use Petros' examples as fixtures — then TDD to refine our implementation to work with them, then (2) add path-based exceptions, then (3) just enable multiline-mode in our RegExp ([using the `m` option](https://ruby-doc.org/3.2.2/Regexp.html#class-Regexp-label-Options)) — instead of investing into a more complex parser — would be enough for the first iteration. I think making our RegEx support multiline might not end up be _that_ complicated after all.

ParaskP7:
> Thanks for your suggestions @ParaskP7! Very helpful specially for us to get a bit of "reality check" on what we can expect.

Thank YOU @iangmaia ! 🥇 

> We can use several of your examples for exceptions.

🤞 

> For example, a host project could just ignore src/main/java/org/wordpress/android/ui/** instead of having to specify many view naming and base class exceptions.

Just an FYI that you should be careful about that because, although this `**/ui/**` package and path seems that is should ONLY contain UI related classes, which anyway can't be unit test, this is actually not the case. For example, every such `ui` sub-package, like `**/ui/about/**` has a `*ViewModel` class, which should be definitely unit tested. Actually, if we only had to pick one type of classes to be unit tested based on the tooling we are developing atm, IMHO then those `*ViewModel` related classed would be at the top of our list.

> Yeah I think that (1) use Petros' examples as fixtures — then TDD to refine our implementation to work with them, then (2) add path-based exceptions, then (3) just enable multiline-mode in our RegExp ([using the m option](https://ruby-doc.org/3.2.2/Regexp.html#class-Regexp-label-Options)) — instead of investing into a more complex parser — would be enough for the first iteration.

👍 

> I think making our RegEx support multiline might not end up be that complicated after all.

🤞

iangmaia:
Hi @ParaskP7 and @AliSoftware!
I was on and off into this and changed my approach a couple of times 🙃 but I have some improvements to share:
- I've first added some code to work with file fixtures, so that we could load external files to use in the tests: 54b23f68b2f0482ecd2a1318fb1d966747e202bd.
- I've created files for everything that I was going to use in the tests, also to improve the tests readability and to make it easier to write more complex cases. Along the way, I also copied a couple of complex class declarations from WPA: 45cfd64491f8a7e658cf73f349b69b85141acf9b (adding the fixture files) and b713e7440de7121ddc845cb4bf59bec55b1f6b06 (using the files in the tests).
- I've made some improvements to the plugin as some tests were failing. Not only the regexes were not supporting multiline declarations as we discussed, but the way it worked before was based on reading the diff line-by-line, so any declaration with an inheritance clause on a second line, for example (like it happens [here](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/WordPress/src/main/java/org/wordpress/android/util/config/BloggingPromptsFeatureConfig.kt)), would have the subclass exception failing. So I changed the logic to use the entire diff: aefb8f7033e0b2e834eff5696029bfa91f0778e5.
- I've added path exceptions here (which will provide another layer to ignore files, such as `*Extension.kt`, for example): ea55caae149e694a7e930e56cf2355e2b63f0493.

Please let me know if you have any comments.

ParaskP7:
👋 @iangmaia !

> I was on and off into this and changed my approach a couple of times 🙃 but I have some improvements to share:

I took a quick look on all those improvements, especially the `test fixture files` added, their usage from within the `unit test checker spec`, and everything LGTM, great improvements from where I stand. ❤️ 

> I've first added some code to work with file fixtures, so that we could load external files to use in the tests: https://github.com/Automattic/dangermattic/commit/54b23f68b2f0482ecd2a1318fb1d966747e202bd.

👍 

> I've created files for everything that I was going to use in the tests, also to improve the tests readability and to make it easier to write more complex cases. Along the way, I also copied a couple of complex class declarations from WPA: https://github.com/Automattic/dangermattic/commit/45cfd64491f8a7e658cf73f349b69b85141acf9b (adding the fixture files) and https://github.com/Automattic/dangermattic/commit/b713e7440de7121ddc845cb4bf59bec55b1f6b06 (using the files in the tests).

❤️ 

> I've made some improvements to the plugin as some tests were failing. Not only the regexes were not supporting multiline declarations as we discussed, but the way it worked before was based on reading the diff line-by-line, so any declaration with an inheritance clause on a second line, for example (like it happens [here](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/WordPress/src/main/java/org/wordpress/android/util/config/BloggingPromptsFeatureConfig.kt)), would have the subclass exception failing. So I changed the logic to use the entire diff: https://github.com/Automattic/dangermattic/commit/aefb8f7033e0b2e834eff5696029bfa91f0778e5.

I am not sure what and if I am doing anything wrong (trying to get ahead of myself), but me trying to follow the [README.md](https://github.com/Automattic/dangermattic/#readme) and run the tests using `bundle exec rake spec`, I get it failing due to the below errors:

<details>
    <summary>Failing due to errors</summary>

```bundle
/Users/.../.rbenv/versions/3.2.2/bin/ruby -I/Users/.../.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/rspec-core-3.12.2/lib:/Users/.../.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/rspec-support-3.12.1/lib /Users/.../.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/rspec-core-3.12.2/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
.....................................................

Finished in 0.6293 seconds (files took 0.77614 seconds to load)
53 examples, 0 failures

Running RuboCop...
Inspecting 15 files
...............

15 files inspected, no offenses detected
bundle exec danger plugins lint
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
[warn]: @param tag has unknown parameter name: Regexes 
    in file `/Users/.../dangermattic/lib/dangermattic/plugins/android_unit_test_checker.rb' near line 161
[warn]: @param tag has unknown parameter name: Regexes 
    in file `/Users/.../dangermattic/lib/dangermattic/plugins/android_unit_test_checker.rb' near line 161

[!] Failed


Errors
  - Examples - AndroidUnitTestChecker (class):
    - You should include some examples of common use-cases for your plugin.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L8#-L27
    - /Users/.../dangermattic/lib/dangermattic/plugins/android_unit_test_checker.rb:7

  - Examples - ManifestPRChecker (class):
    - You should include some examples of common use-cases for your plugin.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L8#-L27
    - /Users/.../dangermattic/lib/dangermattic/plugins/manifest_pr_checker.rb:5

  - Examples - ViewChangesNeedScreenshots (class):
    - You should include some examples of common use-cases for your plugin.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L8#-L27
    - /Users/.../dangermattic/lib/dangermattic/plugins/view_changes_need_screenshots.rb:5


Warnings
  - Tags - AndroidUnitTestChecker (class):
    - This plugin does not include `@tags tag1, tag2` and thus will be harder to find in search.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L30
    - /Users/.../dangermattic/lib/dangermattic/plugins/android_unit_test_checker.rb:7

  - References - AndroidUnitTestChecker (class):
    - Ideally, you have a reference implementation of your plugin that you can show to people, add `@see org/repo` to have the site auto link it.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L29
    - /Users/.../dangermattic/lib/dangermattic/plugins/android_unit_test_checker.rb:7

  - Tags - ManifestPRChecker (class):
    - This plugin does not include `@tags tag1, tag2` and thus will be harder to find in search.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L30
    - /Users/.../dangermattic/lib/dangermattic/plugins/manifest_pr_checker.rb:5

  - References - ManifestPRChecker (class):
    - Ideally, you have a reference implementation of your plugin that you can show to people, add `@see org/repo` to have the site auto link it.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L29
    - /Users/.../dangermattic/lib/dangermattic/plugins/manifest_pr_checker.rb:5

  - Return Type - check_all_manifest_lock_updated (method):
    - If the function has no useful return value, use ` @return  [void]` - this will be ignored by documentation generators.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L46
    - /Users/.../dangermattic/lib/dangermattic/plugins/manifest_pr_checker.rb:8

  - Return Type - check_gemfile_lock_updated (method):
    - If the function has no useful return value, use ` @return  [void]` - this will be ignored by documentation generators.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L46
    - /Users/.../dangermattic/lib/dangermattic/plugins/manifest_pr_checker.rb:15

  - Return Type - check_podfile_lock_updated (method):
    - If the function has no useful return value, use ` @return  [void]` - this will be ignored by documentation generators.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L46
    - /Users/.../dangermattic/lib/dangermattic/plugins/manifest_pr_checker.rb:24

  - Return Type - check_swift_package_resolved_updated (method):
    - If the function has no useful return value, use ` @return  [void]` - this will be ignored by documentation generators.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L46
    - /Users/.../dangermattic/lib/dangermattic/plugins/manifest_pr_checker.rb:33

  - Tags - ViewChangesNeedScreenshots (class):
    - This plugin does not include `@tags tag1, tag2` and thus will be harder to find in search.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L30
    - /Users/.../dangermattic/lib/dangermattic/plugins/view_changes_need_screenshots.rb:5

  - References - ViewChangesNeedScreenshots (class):
    - Ideally, you have a reference implementation of your plugin that you can show to people, add `@see org/repo` to have the site auto link it.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L29
    - /Users/.../dangermattic/lib/dangermattic/plugins/view_changes_need_screenshots.rb:5

  - Return Type - view_changes_need_screenshots (method):
    - If the function has no useful return value, use ` @return  [void]` - this will be ignored by documentation generators.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L46
    - /Users/.../dangermattic/lib/dangermattic/plugins/view_changes_need_screenshots.rb:11

Failing due to errors

rake aborted!
Command failed with status (1): [bundle exec danger plugins lint...]
/Users/.../dangermattic/Rakefile:23:in `block in <top (required)>'
/Users/.../dangermattic/Rakefile:15:in `block in <top (required)>'
/Users/.../.rbenv/versions/3.2.2/bin/bundle:25:in `load'
/Users/.../.rbenv/versions/3.2.2/bin/bundle:25:in `<main>'
Tasks: TOP => spec_docs
(See full trace by running task with --trace)
```

</details>

Kindly advise, I am pretty sure I am doing something wrong here, not being comfortable with ruby and all... 🙏 😊 🙏 

> I've added path exceptions here (which will provide another layer to ignore files, such as *Extension.kt, for example): https://github.com/Automattic/dangermattic/commit/ea55caae149e694a7e930e56cf2355e2b63f0493.

👍

iangmaia:
> I am not sure what and if I am doing anything wrong (trying to get ahead of myself), but me trying to follow the [README.md](https://github.com/Automattic/dangermattic/#readme) and run the tests using `bundle exec rake spec`, I get it failing due to the below errors:
> 
> Failing due to errors
> Kindly advise, I am pretty sure I am doing something wrong here, not being comfortable with ruby and all... 🙏 😊 🙏

Hi @ParaskP7, thanks for this, it's also a reminder for me to review the plugins against Danger's own lint checks (which are also triggered when you run `bundle exec rake spec`) that have a few additional requirements for publishing plugins in their index, such as providing plugin usage examples in the docs; this is something I'll also review when better structuring the documentation for this repo.
In your logs, you probably ran the tests already before these errors/warnings. You can run the tests alone by calling `bundle exec rake specs` or simply `bundle exec rspec`.

ParaskP7:
> Hi @ParaskP7, thanks for this, it's also a reminder for me to review the plugins against Danger's own lint checks (which are also triggered when you run bundle exec rake spec) that have a few additional requirements for publishing plugins in their index, such as providing plugin usage examples in the docs; this is something I'll also review when better structuring the documentation for this repo.

😊 🥇 🙇 

> In your logs, you probably ran the tests already before these errors/warnings.

Yea, I had a feeling about that, seeing this below:

```
Finished in 0.70939 seconds (files took 0.34692 seconds to load)
53 examples, 0 failures
```

But, I wanted to make sure of that, having limited knowledge in this area... 😊

> You can run the tests alone by calling bundle exec rake specs or simply bundle exec rspec.

This is awesome, didn't know you could also do that, thanks a lot for sharing @iangmaia ! ❤️

iangmaia:
Hi @AliSoftware -- have you had the chance to look into [my latest changes](https://github.com/Automattic/dangermattic/pull/1#issuecomment-1662595925)? Please let me know so that we can wrap up this initial version of the plugin. Thanks!

AliSoftware:
Sorry @iangmaia  I've been focused on other projects lately and I know I'm a bit behind on my pending PR reviews of this repo. I'll try to take a look at this later this week though.

lib/dangermattic/plugins/android_unit_test_checker.rb
```diff
@@ -0,0 +1,189 @@
...
+    # Finds added classes that potentially will require a test (violations) in the given file based on the changes in the diff patch.
+    #
+    # @param path [String] The file in the diff to check for violations.
+    # @param diff_patch [String] The diff patch containing the changes to the file.
+    # @param classes_exceptions [Array<String>] An array of class names that are exceptions and should be ignored.
+    # @param subclasses_exceptions [Array<String>] An array of class names whose subclasses should be ignored as well.
+    #
+    # @return [Array<ClassViolation>] An array of ClassViolation objects representing the violations found.
+    def find_violations(path:, diff_patch:, classes_exceptions:, subclasses_exceptions:)
+      added_lines = GitUtils.added_lines(diff_patch: diff_patch)
```
AliSoftware, line 146: I wonder if we shouldn't also include from the diff the lines that are there only for surrounding context, i.e. lines that are neither prefixed with `+` (added) nor with `-` (removed).
That way this could help have a bit more context with the 2–3 previous and next lines around a diff addition, reducing the risk of false positives when the change is about adding or changing a line on an existing declaration for which e.g. the inheritance clause would be on a next (and already previously existing) line.

wdyt?

lib/dangermattic/plugins/utils/git_utils.rb
```diff
@@ -0,0 +1,27 @@
...
+    select_lines(diff_patch: diff_patch, change_type: :removed)
+  end
+
+  def self.change_type(diff_line:)
+    if diff_line.start_with?('+') && !diff_line.start_with?('+++ ')
+      :added
+    elsif diff_line.start_with?('-') && !diff_line.start_with?('--- ')
+      :removed
+    else
+      :other
```
AliSoftware, line 19: If we decide to include non-changed lines like I suggest in the other comment, we might want to differentiate between header-like lines (e.g. starting with `diff --git a/… b/…` or `index …` or `+++`, `---`, `@@ … @@`) and context lines (i.e. starting with a space) 🤔

spec/android_unit_test_checker_spec.rb
```diff
@@ -0,0 +1,333 @@
...
+
+        diff = generate_delete_diff_from_fixtures(deleted_files)
+        allow(@dangerfile.git).to receive(:diff).and_return(diff)
+
+        @plugin.check_missing_tests
+
+        expect(@dangerfile.status_report[:errors]).to be_empty
+      end
+
+      it 'report errors when we remove test classes for classes we refactored' do
```
AliSoftware:
```suggestion
      it 'reports errors when we remove test classes for classes we refactored' do
```

spec/android_unit_test_checker_spec.rb
```diff
@@ -0,0 +1,333 @@
...
+          TestsINeedThem2.kt
+        ]
+
+        diff = generate_add_diff_from_fixtures(added_files)
+        allow(@dangerfile.git).to receive(:diff).and_return(diff)
+        allow(@plugin.github).to receive(:pr_labels).and_return(['unit-tests-exemption'])
+
+        @plugin.check_missing_tests
+
+        expect(@dangerfile.status_report[:errors]).to be_empty
```
AliSoftware: I think we should add an additional `expect` here to check that it reports warnings, though, like the "Class `#{c.classname}` is missing tests, but `#{bypass_label}` label was set to ignore this." message

spec/android_unit_test_checker_spec.rb
```diff
@@ -0,0 +1,333 @@
...
+
+        ignore_label = 'ignore-no-tests'
+
+        diff = generate_add_diff_from_fixtures(added_files)
+        allow(@dangerfile.git).to receive(:diff).and_return(diff)
+        allow(@plugin.github).to receive(:pr_labels).and_return([ignore_label])
+
+        @plugin.check_missing_tests(bypass_label: ignore_label)
+
+        expect(@dangerfile.status_report[:errors]).to be_empty
```
AliSoftware: Same here, probably worth adding another `expect` to check that the `status_report` contains the expected _warnings_

spec/fixtures/android_unit_test_checker/src/android/java/org/activities/MyActivity.kt
```diff
@@ -0,0 +1,7 @@
+import android.content.Context
+import android.content.res.Configuration
+import androidx.appcompat.app.AppCompatActivity
+
+class MyActivity: AppCompatActivity {
+    fun myActivity() {}
+}
```
AliSoftware: Nitpick: add newline at end of file.

spec/fixtures/android_unit_test_checker/src/android/java/org/fragments/MyFragment.kt
```diff
@@ -0,0 +1,6 @@
+import androidx.fragment.app.Fragment
+
+@AndroidEntryPoint
+class MyFragment : Fragment(R.layout.activity_log_item_detail) {
+    override fun onBackPressed() {}
+}
```
AliSoftware: Nitpick: add newline at end of file.

spec/fixtures/android_unit_test_checker/src/androidTest/java/org/test/ToolTest.kt
```diff
@@ -0,0 +1,5 @@
+class ToolTest {
+    fun testMethod() {
+        
+    }
+}
```
AliSoftware: Nitpick: add newline at end of file.

AliSoftware (approved):
Left a couple of suggestions, but still approving because those are not blocking for a first version and could also be fixed in a separate PR if you prefer to land this one and iterate later.

spec/android_unit_test_checker_spec.rb
iangmaia: Nice catch -- replaced it with a custom matcher from another PR that will check that the error report is empty as well as the warnings ba3c549caeba77eeb22fe432b4309b064bcc1504.
I'll review all the plugins later to use this match more broadly.

lib/dangermattic/plugins/android_unit_test_checker.rb
iangmaia: That's a good point -- I've created a task for this (I'm adding all pending issues and suggestions to this project) and will revisit it, before tackling the iOS counterpart, in a separate PR to keep things clean: https://github.com/orgs/Automattic/projects/775/views/1?pane=issue&itemId=37634872

PR #2, opened 2023-06-29
Add Danger plugin to check if changes on view code also have screenshots

iangmaia:
This PR adds a Danger plugin to check whether Pull Requests with changes on UI code for iOS or Android also have corresponding screenshots.

### DIFF
--- a/lib/dangermattic/plugins/view_changes_need_screenshots.rb
+++ b/lib/dangermattic/plugins/view_changes_need_screenshots.rb
@@ -3,8 +3,8 @@
 module Danger
   # Plugin to detect View files in a PR but without having accompanying screenshots.
   class ViewChangesNeedScreenshots < Plugin
-    VIEW_EXTENSIONS_IOS = /(View|Button)\.(swift|m)$|\.xib$|\.storyboard$/.freeze
-    VIEW_EXTENSIONS_ANDROID = /(?i)(View|Button)\.(java|kt|xml)$/.freeze
+    VIEW_EXTENSIONS_IOS = /(View|Button)\.(swift|m)$|\.xib$|\.storyboard$/
+    VIEW_EXTENSIONS_ANDROID = /(?i)(View|Button)\.(java|kt|xml)$/
 
     # Checks if view files have been modified and if a screenshot is included in the pull request body,
     # displaying a warning if view files have been modified but no screenshot is included.
--- a/lib/dangermattic/plugins/view_changes_need_screenshots.rb
+++ b/lib/dangermattic/plugins/view_changes_need_screenshots.rb
@@ -2,7 +2,7 @@
 
 module Danger
   # Plugin to detect View files in a PR but without having accompanying screenshots.
-  class ViewCodePRChecker < Plugin
+  class ViewChangesNeedScreenshots < Plugin
     VIEW_EXTENSIONS_IOS = /(View|Button)\.(swift|m)$|\.xib$|\.storyboard$/.freeze
     VIEW_EXTENSIONS_ANDROID = /(?i)(View|Button)\.(java|kt|xml)$/.freeze
 
--- a/lib/dangermattic/plugins/view_changes_need_screenshots.rb
+++ b/lib/dangermattic/plugins/view_changes_need_screenshots.rb
@@ -15,8 +15,8 @@ module Danger
       pr_has_screenshots = github.pr_body =~ %r{https?://\S*\.(gif|jpg|jpeg|png|svg)}
       pr_has_screenshots ||= github.pr_body =~ /!\[(.*?)\]\((.*?)\)/
 
-      warning = 'View files have been modified, but no screenshot is included in the pull request.'\
-                ' Consider adding one for clarity.'
+      warning = 'View files have been modified, but no screenshot is included in the pull request.' \
+                ' Consider adding some for clarity.'
       warn(warning) if view_files_modified && !pr_has_screenshots
     end
   end
--- a/lib/dangermattic/plugins/view_code_pr_checker.rb
+++ b/lib/dangermattic/plugins/view_code_pr_checker.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin to detect View files in a PR but without having accompanying screenshots.
+  class ViewCodePRChecker < Plugin
+    VIEW_EXTENSIONS_IOS = /(View|Button)\.swift$|\.xib$|\.storyboard$/.freeze
+    VIEW_EXTENSIONS_ANDROID = /(?i)(View|Button)\.(java|kt|xml)$/.freeze
+
+    # Checks if view files have been modified and if a screenshot is included in the pull request body,
+    # displaying a warning if view files have been modified but no screenshot is included.
+    def view_changes_need_screenshots
+      view_files_modified = git.modified_files.any? do |file|
+        VIEW_EXTENSIONS_IOS =~ file || VIEW_EXTENSIONS_ANDROID =~ file
+      end
+      pr_has_screenshots = github.pr_body =~ %r{https?://\S*\.(gif|jpg|jpeg|png|svg){1}}
+      pr_has_screenshots ||= github.pr_body =~ /!\[(.*?)\]\((.*?)\)/
+
+      warning = 'View files have been modified, but no screenshot is included in the pull request.'\
+                ' Consider adding one for clarity.'
+      warn(warning) if view_files_modified && !pr_has_screenshots
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


lib/dangermattic/plugins/view_code_pr_checker.rb
```diff
@@ -0,0 +1,23 @@
...
+    VIEW_EXTENSIONS_IOS = /(View|Button)\.swift$|\.xib$|\.storyboard$/.freeze
+    VIEW_EXTENSIONS_ANDROID = /(?i)(View|Button)\.(java|kt|xml)$/.freeze
+
+    # Checks if view files have been modified and if a screenshot is included in the pull request body,
+    # displaying a warning if view files have been modified but no screenshot is included.
+    def view_changes_need_screenshots
+      view_files_modified = git.modified_files.any? do |file|
+        VIEW_EXTENSIONS_IOS =~ file || VIEW_EXTENSIONS_ANDROID =~ file
+      end
+      pr_has_screenshots = github.pr_body =~ %r{https?://\S*\.(gif|jpg|jpeg|png|svg){1}}
```
AliSoftware: I'm not sure what's the point of the `{1}` at the end of this regex? It's basically saying "repeat the previous token exactly 1 time" which should be the same as not having the `{1}` repeater pattern at all 🙃

spec/view_code_pr_checker_spec.rb
AliSoftware, line 1: I see that for each case in that file, you have a "warns when a PR …" for iOS then a `warns when an Android PR …` corresponding test.

 - For consistency, you should include "iOS" in the first case, as there's no reason to mention "Android" in the second case but not mention the platform (iOS) in the first case.
 - Also, I'd suggest to maybe use `describe` or `context` to group similar tests togethers. Or to group all iOS tests together, then all Android tests together. I think the even better approach would be to use `shared_examples` to provide the 4 use cases you're testing, then call those shared examples once with iOS file names, once with Android file names, or something like that. In other words, DRY and mutualize some code.

lib/dangermattic/plugins/view_code_pr_checker.rb
iangmaia: I guess I tried something else in the regex and forgot to remove this 🙃

spec/view_code_pr_checker_spec.rb
iangmaia: >  Or to group all iOS tests together, then all Android tests together. I think the even better approach would be to use shared_examples to provide the 4 use cases you're testing, then call those shared examples once with iOS file names, once with Android file names, or something like that. In other words, DRY and mutualize some code.

Nice suggestion! Will change it.

lib/dangermattic/plugins/view_code_pr_checker.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin to detect View files in a PR but without having accompanying screenshots.
+  class ViewCodePRChecker < Plugin
```
AliSoftware: I wonder if we couldn't pick a more explicit name for that plugin. Just reading `ViewCodePRChecker` feels like it could equally be a plugin that allows you to "View" information about your "Code" in your "PR" (like stats about code changes and stuff).

Maybe `ViewChangesNeedScreenshots`, or `ViewRelatedChangesHasScreenshots < Plugin`, or something like that?

spec/view_code_pr_checker_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      it 'warns when a PR with view code changes does not have screenshots' do
+        allow(@plugin.github).to receive(:pr_body).and_return('PR Body')
+
+        @plugin.view_changes_need_screenshots
+
+        expect(@dangerfile.status_report[:warnings].count).to eq 1
+      end
+
+      it 'does nothing when a PR with view code changes has screenshots' do
```
AliSoftware: You put that shared example about "has view code and also **has** screenshots" inside the `shared_examples` that is named "view code **without** screenshots". This seems contradictory.

The easiest fix for that is probably to rename the parent `shared_examples` to just `'PR has view code'`.

spec/view_code_pr_checker_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+require_relative 'spec_helper'
+
+module Danger
+  describe Danger::ViewCodePRChecker do
+    it 'should be a plugin' do
+      expect(described_class.new(nil)).to be_a Danger::Plugin
+    end
+
+    shared_examples 'view code without screenshots' do |modified_files|
```
AliSoftware:
```suggestion
    shared_examples 'PR with view code changes' do |modified_files|
```

spec/view_code_pr_checker_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        @plugin = @dangerfile.view_code_pr_checker
+      end
+
+      context 'iOS PR has view code changes' do
+        include_examples 'view code without screenshots', ['TestView.swift', 'SimpleViewHelper.m']
+      end
+
+      context 'iOS PR has button changes' do
+        include_examples 'view code without screenshots', ['SimpleViewHelper.m', 'MyAwesomeButton.swift']
+      end
```
AliSoftware: Now that we have those nice `shared_examples` that makes it a one-liner to create new tests, I think it would be worth it, all while not adding too much, to duplicate those two tests but with ObjC.

```suggestion
      context 'iOS PR has view code changes in Swift files' do
        include_examples 'PR with view code changes', ['TestView.swift', 'SimpleViewHelper.m']
      end

      context 'iOS PR has button changes in Swift files' do
        include_examples 'PR with view code changes', ['SimpleViewHelper.m', 'MyAwesomeButton.swift']
      end

      context 'iOS PR has view code changes in ObjC files' do
        include_examples 'PR with view code changes', ['TestView.m', 'SimpleViewHelper.m']
      end

      context 'iOS PR has button changes in ObjC files' do
        include_examples 'PR with view code changes', ['SimpleViewHelper.m', 'MyAwesomeButton.m']
      end
```
And similar for Java vs Kotlin below of course.

AliSoftware: Actually now that I think about it, maybe it would be cleaner to have:
```ruby
shared_examples 'view_code_pr_checker` do
  it 'warns when a PR with view code changes does not have screenshots' do … end
  it 'does nothing when a PR with view code changes has screenshots' do … end
  it 'does nothing when a PR does not have view code changes' do … end
end

context 'iOS PR with Swift files' do
  include_examples 'view_code_pr_checker', ['SimpleViewHelper.swift', 'MyButtonTester.swift']
end

context 'iOS PR with ObjC files' do
  include_examples 'view_code_pr_checker', ['SimpleViewHelper.m', 'MyButtonTester.m']
end

context 'Android PR with Kotlin files' do
  include_examples 'view_code_pr_checker', ['test_view.xml', 'SimpleViewHelper.kt']
end

context 'Android PR with Java files' do
  include_examples 'view_code_pr_checker', ['SimpleViewHelper.java', 'MyAwesomeButton.java']
end
```
?

lib/dangermattic/plugins/view_code_pr_checker.rb
iangmaia: Great suggestion, yesterday when reviewing the code I also thought the same. `ViewChangesNeedScreenshots` sounds good to me!

spec/view_code_pr_checker_spec.rb
iangmaia: The problem grouping everything in one single shared example is that the `PR without view code` check will break when you DO have view code being changed, as warning will be generated. That's why I had to separate the examples.

But I like the idea of adding separate ObjC/Java tests!

lib/dangermattic/plugins/view_changes_need_screenshots.rb
```diff
@@ -0,0 +1,23 @@
...
+    # displaying a warning if view files have been modified but no screenshot is included.
+    def view_changes_need_screenshots
+      view_files_modified = git.modified_files.any? do |file|
+        VIEW_EXTENSIONS_IOS =~ file || VIEW_EXTENSIONS_ANDROID =~ file
+      end
+      pr_has_screenshots = github.pr_body =~ %r{https?://\S*\.(gif|jpg|jpeg|png|svg)}
+      pr_has_screenshots ||= github.pr_body =~ /!\[(.*?)\]\((.*?)\)/
+
+      warning = 'View files have been modified, but no screenshot is included in the pull request.'\
+                ' Consider adding one for clarity.'
```
AliSoftware:
```suggestion
      warning = 'View files have been modified, but no screenshot is included in the pull request.' \
                ' Consider adding some for clarity.'
```

AliSoftware (approved):

PR #3, opened 2023-06-29
Add Danger plugin to check if a manifest change has a corresponding lock change

iangmaia:
This PR adds a Danger plugin to check if a Pull Request containing a change in a manifest file such as `Gemfile` or `Package.swift` have a corresponding change on a lock file (`Gemfile.lock` or `Package.resolved`).

### DIFF
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -13,6 +13,15 @@ module Danger
       )
     end
 
+      # Check if the `Podfile` file was modified without a corresponding `Podfile.lock` update 
+      def check_podfile_lock_updated
+        check_manifest_lock_updated(
+          file_name: 'Podfile',
+          lock_file_name: 'Podfile.lock',
+          instruction: 'Please run `bundle exec pod install`'
+        )
+      end
+    
     # Check if the `Package.swift` file was modified without a corresponding `Package.resolved` update 
     def check_swift_package_resolved_updated
       check_manifest_lock_updated(
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -38,7 +38,7 @@ module Danger
       lock_modified = git.modified_files.include?(lock_file_name)
       
       if manifest_modified && !lock_modified
-        failure("#{file_name} was changed without updating #{lock_file_name}. #{instruction}.")
+        warn("#{file_name} was changed without updating #{lock_file_name}. #{instruction}.")
       end
     end
   end
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -3,6 +3,15 @@
 module Danger
   # Plugin to check if the Gemfile.lock was updated when changing the Gemfile in a PR.
   class ManifestPRChecker < Plugin
+
+    # Performs all the checks, asserting that changes on `Gemfile`, `Podfile` and `Package.swift` must have corresponding
+    # lock file changes.
+    def check_all_manifest_lock_updated
+      check_gemfile_lock_updated
+      check_podfile_lock_updated
+      check_swift_package_resolved_updated
+    end
+
     # Check if the `Gemfile` file was modified without a corresponding `Gemfile.lock` update
     def check_gemfile_lock_updated
       check_manifest_lock_updated(
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -3,7 +3,6 @@
 module Danger
   # Plugin to check if the Gemfile.lock was updated when changing the Gemfile in a PR.
   class ManifestPRChecker < Plugin
-
     # Performs all the checks, asserting that changes on `Gemfile`, `Podfile` and `Package.swift` must have corresponding
     # lock file changes.
     def check_all_manifest_lock_updated
@@ -42,10 +41,10 @@ module Danger
     private
 
     def check_manifest_lock_updated(file_name:, lock_file_name:, instruction:)
-      manifest_modified = git.modified_files.include?(file_name)
-      lock_modified = git.modified_files.include?(lock_file_name)
+      manifest_modified_file = git.modified_files.find { |f| f.end_with?(file_name) }
+      lock_modified = git.modified_files.any? { |f| f.end_with?(lock_file_name) && manifest_modified_file && File.dirname(f) == File.dirname(manifest_modified_file) }
 
-      return unless manifest_modified && !lock_modified
+      return unless manifest_modified_file && !lock_modified
 
       warn("#{file_name} was changed without updating #{lock_file_name}. #{instruction}.")
     end
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -41,12 +41,16 @@ module Danger
     private
 
     def check_manifest_lock_updated(file_name:, lock_file_name:, instruction:)
-      manifest_modified_file = git.modified_files.find { |f| f.end_with?(file_name) }
-      lock_modified = git.modified_files.any? { |f| f.end_with?(lock_file_name) && manifest_modified_file && File.dirname(f) == File.dirname(manifest_modified_file) }
+      # Find all the modified manifest files
+      manifest_modified_files = git.modified_files.select { |f| File.basename(f) == file_name }
 
-      return unless manifest_modified_file && !lock_modified
+      # For each manifest file, check if the corresponding lockfile (in the same dir) was also modified
+      manifest_modified_files.each do |manifest_file|
+        lockfile_modified = git.modified_files.any? { |f| File.dirname(f) == File.dirname(manifest_file) && File.basename(f) == lock_file_name }
+        next if lockfile_modified
 
-      warn("#{file_name} was changed without updating #{lock_file_name}. #{instruction}.")
+        warn("#{manifest_file} was changed without updating its corresponding #{lock_file_name}. #{instruction}.")
+      end
     end
   end
 end
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -3,8 +3,7 @@
 module Danger
   # Plugin to check if the Gemfile.lock was updated when changing the Gemfile in a PR.
   class ManifestPRChecker < Plugin
-
-    # Check if the `Gemfile` file was modified without a corresponding `Gemfile.lock` update 
+    # Check if the `Gemfile` file was modified without a corresponding `Gemfile.lock` update
     def check_gemfile_lock_updated
       check_manifest_lock_updated(
         file_name: 'Gemfile',
@@ -13,16 +12,16 @@ module Danger
       )
     end
 
-      # Check if the `Podfile` file was modified without a corresponding `Podfile.lock` update 
-      def check_podfile_lock_updated
-        check_manifest_lock_updated(
-          file_name: 'Podfile',
-          lock_file_name: 'Podfile.lock',
-          instruction: 'Please run `bundle exec pod install`'
-        )
-      end
-    
-    # Check if the `Package.swift` file was modified without a corresponding `Package.resolved` update 
+    # Check if the `Podfile` file was modified without a corresponding `Podfile.lock` update
+    def check_podfile_lock_updated
+      check_manifest_lock_updated(
+        file_name: 'Podfile',
+        lock_file_name: 'Podfile.lock',
+        instruction: 'Please run `bundle exec pod install`'
+      )
+    end
+
+    # Check if the `Package.swift` file was modified without a corresponding `Package.resolved` update
     def check_swift_package_resolved_updated
       check_manifest_lock_updated(
         file_name: 'Package.swift',
@@ -36,10 +35,10 @@ module Danger
     def check_manifest_lock_updated(file_name:, lock_file_name:, instruction:)
       manifest_modified = git.modified_files.include?(file_name)
       lock_modified = git.modified_files.include?(lock_file_name)
-      
-      if manifest_modified && !lock_modified
-        warn("#{file_name} was changed without updating #{lock_file_name}. #{instruction}.")
-      end
+
+      return unless manifest_modified && !lock_modified
+
+      warn("#{file_name} was changed without updating #{lock_file_name}. #{instruction}.")
     end
   end
 end
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin to check if the Gemfile.lock was updated when changing the Gemfile in a PR.
+  class ManifestPRChecker < Plugin
+
+    # Check if the `Gemfile` file was modified without a corresponding `Gemfile.lock` update 
+    def check_gemfile_lock_updated
+      check_manifest_lock_updated(
+        file_name: 'Gemfile',
+        lock_file_name: 'Gemfile.lock',
+        instruction: 'Please run `bundle install` or `bundle update <updated_gem>`'
+      )
+    end
+
+    # Check if the `Package.swift` file was modified without a corresponding `Package.resolved` update 
+    def check_swift_package_resolved_updated
+      check_manifest_lock_updated(
+        file_name: 'Package.swift',
+        lock_file_name: 'Package.resolved',
+        instruction: 'Please resolve the Swift packages in Xcode'
+      )
+    end
+
+    private
+
+    def check_manifest_lock_updated(file_name:, lock_file_name:, instruction:)
+      manifest_modified = git.modified_files.include?(file_name)
+      lock_modified = git.modified_files.include?(lock_file_name)
+      
+      if manifest_modified && !lock_modified
+        failure("#{file_name} was changed without updating #{lock_file_name}. #{instruction}.")
+      end
+    end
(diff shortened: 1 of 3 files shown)


spec/manifest_pr_checker_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require_relative 'spec_helper'
+
+module Danger
+  describe Danger::ManifestPRChecker do
+    it 'should be a plugin' do
+      expect(Danger::ManifestPRChecker.new(nil)).to be_a Danger::Plugin
+    end
```
mokagio: Once the three plugins (this plus #1 and #2) are all on `trunk`, this check could be made into a shared example or utility method using [`described_class`](https://apidock.com/rspec/Spec/Example/ExampleGroupMethods/described_class).

I haven't tried it, but I think this should work

```suggestion
    it 'should be a plugin' do
      expect(described_class.new(nil)).to be_a Danger::Plugin
    end
```

spec/manifest_pr_checker_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    describe 'with Dangerfile' do
+      before do
+        @dangerfile = testing_dangerfile
+        @my_plugin = @dangerfile.manifest_pr_checker
+      end
+
+      it 'returns an error when a PR changed the Gemfile but not Gemfile.lock' do
```
mokagio: Nitpick

```suggestion
      it 'returns an error when a PR changed the Gemfile but not the Gemfile.lock' do
```

or:

```suggestion
      it 'returns an error when a PR changed Gemfile but not Gemfile.lock' do
```

~~(_I prefer the first suggestion_)~~ Update: I noticed the other example doesn't use an article for either `Gemfile[.lock]`, so maybe the second is more consistent.

spec/manifest_pr_checker_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+module Danger
+  describe Danger::ManifestPRChecker do
+    it 'should be a plugin' do
+      expect(Danger::ManifestPRChecker.new(nil)).to be_a Danger::Plugin
+    end
+
+    describe 'with Dangerfile' do
+      before do
+        @dangerfile = testing_dangerfile
+        @my_plugin = @dangerfile.manifest_pr_checker
```
mokagio: What do you think of calling this `plugin` or `sut` for conciseness?

mokagio (commented):
Have you considered adding `Podfile` here as well? 

Or, is it not here because [we already check that via our Buildkite plugin](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/29a36bfe20125f728f8d62761c7915af31e89bcb/bin/install_cocoapods#L34-L45)?

If so, have you considered moving these checks to the plugin, too?

I guess there's pros and cons for both options. Having them here make them CI-provider agnostic. Having them as part of CI managing the gems/pods/packages installation/cache resolution saves us the trouble of checking elsewhere.

Note: I did notice that this checks simply whether a file changed without the matching lockfile having changed while CI checks the diff in the lockfile. Still, the feedback we get is quite similar. Although, I think CI could give a more precise feedback because, by running the command that modifies the lockfile, it's then able to print the diff of the changes it found, which can sometimes be helpful for debugging.

spec/manifest_pr_checker_spec.rb
iangmaia: I think I'll just change everything to use an article, to make it consistent. It also sounds better to me 👍

lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    private
+
+    def check_manifest_lock_updated(file_name:, lock_file_name:, instruction:)
+      manifest_modified = git.modified_files.include?(file_name)
+      lock_modified = git.modified_files.include?(lock_file_name)
+      
+      if manifest_modified && !lock_modified
+        failure("#{file_name} was changed without updating #{lock_file_name}. #{instruction}.")
+      end
```
AliSoftware: I'm wondering if there could be cases where the manifest file was modified but the lockfile wouldn't.

For example, if you update your `Package.swift` to change tabs to spaces or fix some indentation or add blank lines for formatting or add/remove comments… I think the generated `Package.resolved` would still end up unmodified?

And it might be the same for `Gemfile.lock`?

(It's not the case for `Podfile.lock` because those contain a `PODFILE CHECKSUM` field at the end of the lockfile, so even if the `Podfile` file changed without changing semantics or the list of pods nor their version reauirements, the `Podfile.lock` would still be modified, at the very least to have that checksum updated)

iangmaia: > I'm wondering if there could be cases where the manifest file was modified but the lockfile wouldn't.

Good point -- I thought about that as well, but while I believe this indeed will be the case, it tends to be somewhat rare. Still annoying to have a false negative validation. A couple of alternatives:
1. Have just a warning instead, which can just be ignored when that happens. It is the simples solution, but in a situation where the files are actually out of sync and the warning is there, it might become _too easy_ to ignore it.
2. Make the change comparison a bit more sophisticated, which is more work to implement (I haven't fully thought it out to consider complexity and all cases, but it is certainly a lot more complex than just checking "both files were changed") but will make the plugin more reliable and correct.

I'm tempted to just go with 1 for now, but I can also try quickly looking into it to have some idea of how complex 2 would be -- perhaps it wouldn't be so complicated after all. For the `Gemfile`, we could probably use something from Bundler such as `Bundler::Definition` and `Bundler::LockfileParser`, but for `Package.swift` I guess it would be a bit more complicated 🤔

iangmaia:
Hey @mokagio, thanks for your comments.

> Or, is it not here because [we already check that via our Buildkite plugin](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/29a36bfe20125f728f8d62761c7915af31e89bcb/bin/install_cocoapods#L34-L45)?

Exactly. It is simple enough to add a Podfile check though, so I might as well just do that, but of course still keeping the CI check.

> If so, have you considered moving these checks to the plugin, too?
> I guess there's pros and cons for both options. Having them here make them CI-provider agnostic. Having them as part of CI managing the gems/pods/packages installation/cache resolution saves us the trouble of checking elsewhere.

I think this should be a long term goal of `Dangermattic`: whenever possible, to aggregate the various checks we're performing in different places.

> Note: I did notice that this checks simply whether a file changed without the matching lockfile having changed while CI checks the diff in the lockfile. Still, the feedback we get is quite similar. Although, I think CI could give a more precise feedback because, by running the command that modifies the lockfile, it's then able to print the diff of the changes it found, which can sometimes be helpful for debugging.

Indeed. This is something that this plugin could also do in a rudimentary way, not resolving versions and transitive dependencies and purely comparing the files' content.

lib/dangermattic/plugins/manifest_pr_checker.rb
iangmaia: Update: I've just changed the code to emit warnings for now. I think we could also consider this version "feature complete", and perhaps it is worth it to brainstorm some ideas for making the check smarter (and more precise) on a second iteration.

lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin to check if the Gemfile.lock was updated when changing the Gemfile in a PR.
+  class ManifestPRChecker < Plugin
```
AliSoftware, line 5: I wonder if it wouldn't be worth also providing a helper method for the call sites, like this:
```ruby
def check_all_manifest_lock_updated
  check_gemfile_lock_updated
  check_podfile_lock_updated
  check_swift_package_resolved_updated
end
```
That would make easier to use for projects that are likely to have them all (and if the project don't have one of them, e.g. they don't use SPM, then `check_manifest_lock_updated` will never raise the warning for those ones since `manifest_modified` will always be `false` for those projects not using `Package.swift` anyway)

wdyt?

spec/manifest_pr_checker_spec.rb
AliSoftware, line 1: It would be nice to group each set of `it` tests that relate to the same type of manifest into their own `describe` groups (`describe 'CocoaPods'`, `describe 'Bundler'`, `describe 'SPM'` for example. Or `describe 'Podfile.lock'`, `describe 'Gemfile.lock'`, `describe 'Package.resolved'`?)

lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    # Check if the `Podfile` file was modified without a corresponding `Podfile.lock` update
+    def check_podfile_lock_updated
+      check_manifest_lock_updated(
+        file_name: 'Podfile',
+        lock_file_name: 'Podfile.lock',
+        instruction: 'Please run `bundle exec pod install`'
+      )
+    end
```
AliSoftware, line 30: While this covers the most common case, there are some repositories which also contains `Podfile`+`Podfile.lock` files that are **not** at the root of the repo.

One example that comes to mind is Tumblr-iOS and its `Components/` sub-projects all hosted in the Tumblr-iOS monorepo.
Another example of such repo would be library (pod) repos which contains a Sample project, where the `MyLib.podspec` would be at the root but there would be a `SampleApp/Podfile` and `SampleApp/Podfile.lock` files there.

Given this, maybe the `check_manifest_lock_updated` helper method should not just check for that exact path at the repo root when looking at `git.modified_files`, but also consider any `git.modified_files` whose paths ends with the `file_name` (and compare it with the file that lives in the same `dirname` but has `lock_file_name` as its `basename`)?

lib/dangermattic/plugins/manifest_pr_checker.rb
iangmaia: Ah, it's simple enough to just do it 🙂 it's also nice to have just a single entry point.

spec/manifest_pr_checker_spec.rb
iangmaia: Yeah I like it, it's nice to group them. `describe 'CocoaPods'`, `describe 'Bundler'`, `describe 'SPM'` feels more **descriptive** 😄

lib/dangermattic/plugins/manifest_pr_checker.rb
iangmaia: Hmm, indeed it might happen. I've slightly changed the code to cater for that.

lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -0,0 +1,52 @@
...
+        file_name: 'Package.swift',
+        lock_file_name: 'Package.resolved',
+        instruction: 'Please resolve the Swift packages in Xcode'
+      )
+    end
+
+    private
+
+    def check_manifest_lock_updated(file_name:, lock_file_name:, instruction:)
+      manifest_modified_file = git.modified_files.find { |f| f.end_with?(file_name) }
```
AliSoftware: This only finds the first manifest file.

<details><summary>What if there are many of them (like for the case of <code>Components/**/Podfile</code> in Tumblr-iOS)?</summary>
image
</details>

AliSoftware: You should check `File.basename(f) == file_name` instead of `f.end_with?(file_name)`. Because, even if it's very unlikely that such a file will exist in the repo, `end_with?('Podfile')` would also pick up `Sample/ExamplePodfile` and similar, not just files whose basename is exactly `Podfile`.

lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -0,0 +1,52 @@
...
+      lock_modified = git.modified_files.any? { |f| f.end_with?(lock_file_name) && manifest_modified_file && File.dirname(f) == File.dirname(manifest_modified_file) }
```
AliSoftware: Same limitation of this only accounting for the case of only having a single occurrence of the manifest+lockfile pair in the whole repo, instead of accounting for the possibility of having multiple `Podfile` & `Podfile.lock` in the same repo (like Tumblr-iOS does)

spec/manifest_pr_checker_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+        it 'returns no warnings when both the Podfile and the Podfile.lock were updated' do
+          modified_files = ['Podfile', 'Podfile.lock']
+          allow(@plugin.git).to receive(:modified_files).and_return(modified_files)
+
+          @plugin.check_podfile_lock_updated
+
+          expect(@dangerfile.status_report[:warnings]).to be_empty
+        end
+
+        it 'returns a warning when a PR changed a custom located Podfile but not the corresponding Podfile.lock' do
```
AliSoftware, line 67: We need another test that tests cases like:

```
modified_files = ['./dir1/Podfile', './dir2/Podfile', './dir1/Podfile.lock']
```
To ensure that while the one in `dir1` found its corresponding lockfile, the one in `dir2` missing its lockfile update is still correctly flagged by the plugin.

lib/dangermattic/plugins/manifest_pr_checker.rb
AliSoftware: Maybe something like this
```ruby
# Find all the modified manifest files
manifest_modified_files = git.modified_files.select { |f| File.basename(f) == file_name }
# For each one, check if the corresponding lockfile (in the same dir) was also modified
manifest_modified_files.each do |manifest_file|
  lockfile_modified = git.modified_files.include? do |f|
    File.dirname(f) == File.dirname(manifest_file) && File.basename(f) == lock_file_name
  end
  next if lockfile_modified

  warn("#{manifest_file} was changed without updating its corresponding #{lock_file_name}. #{instruction}."
end
```

PS: This is just a suggestion, not tested. In particular, we need to make sure (via additional unit tests ideally to cover those cases too) that my `File.join(File.dirname(manifest), lock_file_name)` will still work even if `git.modified_files` contains

spec/manifest_pr_checker_spec.rb
AliSoftware: btw, I'd suggest to use TDD here (aka Red-Green-Refactoring), i.e.:
 - First implement the additional tests for the expected behavior,
 - Then run `rspec` and see the tests go red, proving that the tests you wrote would properly catch such a regression
 - Then update the implementation (e.g. [see if my suggested snippet above](https://github.com/Automattic/dangermattic/pull/3#discussion_r1265804027) would work) to make the test go green
 - Then once the test is green, optionally refactor to optimize the code… if needed.

lib/dangermattic/plugins/manifest_pr_checker.rb
iangmaia: Ah, I got attached to the fact of "custom located Podfile" but of course multiple ones can exist 🙂
Good suggestion, gonna try something similar.

spec/manifest_pr_checker_spec.rb
iangmaia: I've implemented the changes we discussed on Monday, by the way.

spec/manifest_pr_checker_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+          modified_files = ['./path/to/Podfile', './my/Podfile.lock']
+          allow(@plugin.git).to receive(:modified_files).and_return(modified_files)
+
+          @plugin.check_podfile_lock_updated
+
+          expected_warning = ['./path/to/Podfile was changed without updating its corresponding Podfile.lock. Please run `bundle exec pod install`.']
+          expect(@dangerfile.status_report[:warnings]).to eq expected_warning
+        end
+
+        it 'returns multiple warnings when a PR changed a custom located Podfile but not the corresponding Podfile.lock' do
```
AliSoftware:
```suggestion
        it 'returns multiple warnings when a PR changed multiple custom located Podfiles but not the corresponding Podfile.lock' do
```

AliSoftware (approved):
I wonder if we should make this a `failure` for `Podfile.lock` (which contains the sha1 of the `Podfile` as part of the lockfile fields, which means that even if the change in `Podfile` was only to add comments and newline and indentation without changing pods being included, the `Podfile.lock` will always need to be updated — unlike `Gemfile.lock` and `Package.resolved`, for which this test can still find false positives and should thus remain a warning)

On the other hand, I like the consistency of all of the `check_*_lock_updated` methods to all behave the same and return the same type of message, so maybe it's better that way.

iangmaia:
> I wonder if we should make this a `failure` for `Podfile.lock` (which contains the sha1 of the `Podfile` as part of the lockfile fields, which means that even if the change in `Podfile` was only to add comments and newline and indentation without changing pods being included, the `Podfile.lock` will always need to be updated — unlike `Gemfile.lock` and `Package.resolved`, for which this test can still find false positives and should thus remain a warning)
> 
> On the other hand, I like the consistency of all of the `check_*_lock_updated` methods to all behave the same and return the same type of message, so maybe it's better that way.

Yeah, I think we can keep it this way for now for consistency.
When we find a way to make these checks more precise or just get more feedback once this is used, we can revisit the usage of warning vs. failure, even if only specifically for the Podfile check.

mokagio:
> Yeah, I think we can keep it this way for now for consistency.
Fair 👌 

> When we find a way to make these checks more precise or just get more feedback once this is used, we can revisit the usage of warning vs. failure, even if only specifically for the Podfile check.

Or add an option to let the caller decide the severity of the violation, with appropriate defaults 🤓

iangmaia:
> Or add an option to let the caller decide the severity of the violation, with appropriate defaults 🤓

This is a good point @mokagio, and not only for this plugin 👍 In a few other occasions, I had a `fail_on_error:` parameter with a default value, and perhaps it's worth it to review where this is applicable at some point (or perhaps it can make sense to have a common configuration field per plugin instead).
I think this kind of plugin standardisation and common configurations are also going to be better figured out once we have more plugins and actual app usage.

For this plugin, for now, I feel a bit reluctant to even have this option as the checks _might_ yield false positives. I think it makes sense just to be consistent and report warnings as we discussed.

PR #4, opened 2023-06-30
Add `.ruby-version` file with latest Ruby version 3.2.2

mokagio:
_What it says in the title._

I noticed the was no `.ruby-version` file while looking at the recent PRs.

This is a very tiny change, but I thought I'd open a PR instead of just pointing it out to have a chance to decide whether to use Ruby 3.2.2 (latest 3.x), 2.7.4 (latest 2.x and what our apps use), or some other version.

My preference is obviously for 3.2.2 as it's the latest stable. However, I don't know what building on 3.x and running on 2.x would entail.


iangmaia:
> My preference is obviously for 3.2.2 as it's the latest stable. However, I don't know what building on 3.x and running on 2.x would entail.

I'm also not sure on the impact of 3.x on Danger itself and other dependencies, but it's perhaps worth trying -- worst case, we can always revert it to 2.7.4 as the impact shouldn't be that big this early in the project. Wdyt?

iangmaia (approved):

mokagio:
> worst case, we can always revert it to 2.7.4 as the impact shouldn't be that big this early in the project. Wdyt?

Sounds good. Hopefully we'll get to use this gem in one of our apps soon and get real world validation.

PR #6, opened 2023-07-05
Buildkite Pipeline Configuration

iangmaia:
Adds a basic Pipeline configuration for Buildkite, running RuboCop and Unit Tests.

### DIFF
--- a/buildkite/pipeline.yml
+++ b/buildkite/pipeline.yml
@@ -1,12 +1,9 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
-  plugins:
-  - &docker_plugin
-    docker#v3.8.0:
-      image: &ruby_version "ruby:3.2.2"
-      propagate-environment: true
-      environment:
-        - "RUBYGEMS_API_KEY"
+  - &common_plugins
+    - &docker_plugin
+      docker#v3.8.0:
+        image: &ruby_version "ruby:3.2.2"
 
 steps:
   #################
@@ -18,13 +15,11 @@ steps:
       - label: "🧪 Build and Test"
         command: |
           echo "--- :rubygems: Setting up Gems"
-          install_gems
+          bundle install
 
           echo "--- :rspec: Run Rspec"
           bundle exec rspec --profile 10 --format progress
-        plugins:
-          - automattic/a8c-ci-toolkit#2.18.1
-          - *docker_plugin
+        plugins: *common_plugins
         agents:
           queue: "default"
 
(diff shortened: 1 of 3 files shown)


buildkite/pipeline.yml
```diff
@@ -0,0 +1,39 @@
+# Nodes with values to reuse in the pipeline.
+common_params:
+  - &common_plugins
+    - &docker_plugin
+      docker#v3.8.0:
+        image: &ruby_version "ruby:3.2.2"
```
crazytonyli: There are a few times CI jobs failed due to [the Docker Hub rate limit](https://docs.docker.com/docker-hub/download-rate-limit/). We have since started using [AWS Public ECR images](https://gallery.ecr.aws/docker/library/ruby). I'd suggestion that we use `public.ecr.aws/docker/library/ruby:3.2.2` here too.

buildkite/pipeline.yml
```diff
@@ -0,0 +1,39 @@
...
+      - label: "🧪 Build and Test"
+        command: |
+          echo "--- :rubygems: Setting up Gems"
+          bundle install
+
+          echo "--- :rspec: Run Rspec"
+          bundle exec rspec --profile 10 --format progress
+        plugins: *common_plugins
+        agents:
+          queue: "default"
```
crazytonyli: This queue setting is redundant: The `default` queue is used if no `agents.queue` is set.

crazytonyli (changes requested):

buildkite/pipeline.yml
AliSoftware: Tbh, personally I like that we make the agent queue explicit in the pipeline file, also to show that this was a conscious choice (†). I'd be ok to set that `agents: queue: "default"` at the top-level of the `pipeline.yml` file to avoid repeating it on each step though and clean things up in the individual steps. WDYT?

---
_(†) especially saying this because [I'm personally experimenting with some stuff in Tumblr right now, trying to make some of our lint steps run on Linux agents instead of Macs if possible](https://github.tumblr.net/TumblrMobile/orangina/pull/25461), and I had to try different agents and think carefully about which one would make more sense for each step)_

crazytonyli: > I'd be ok to set that agents: queue: "default" at the top-level of the pipeline.yml file to avoid repeating it on each step

Sounds like a good idea 👍

buildkite/pipeline.yml
iangmaia: Good to know, thanks 👍

buildkite/pipeline.yml
iangmaia: > I like that we make the agent queue explicit in the pipeline file, also to show that this was a conscious choice

I agree. I moved the default queue parameter to the top.

crazytonyli (approved):
:shipit:

PR #7, opened 2023-07-12
 Add plugin to check for labels in a PR

iangmaia:
Add plugin to check if a PR has at least one label or block the merge if it has been marked as "DO NOT MERGE".

Based on the Peril rule: https://github.com/Automattic/peril-settings/blob/master/org/pr/label.ts

### DIFF
--- a/lib/danger_plugin.rb
+++ b/lib/danger_plugin.rb
@@ -1,8 +1,3 @@
 # frozen_string_literal: true
 
-require 'dangermattic/plugins/android_unit_test_checker'
-require 'dangermattic/plugins/labels_pr_checker'
-require 'dangermattic/plugins/manifest_pr_checker'
-require 'dangermattic/plugins/milestone_checker'
-require 'dangermattic/plugins/pr_size_checker'
-require 'dangermattic/plugins/view_changes_need_screenshots'
+Dir[File.join(__dir__, 'dangermattic/plugins', '*.rb')].each { |file| require file }
--- a/lib/danger_plugin.rb
+++ b/lib/danger_plugin.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'dangermattic/plugins/android_unit_test_checker'
+require 'dangermattic/plugins/labels_pr_checker'
 require 'dangermattic/plugins/manifest_pr_checker'
 require 'dangermattic/plugins/milestone_checker'
 require 'dangermattic/plugins/pr_size_checker'
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -22,22 +22,22 @@ module Danger
     def check(do_not_merge_labels: DEFAULT_DO_NOT_MERGE_LABELS, required_labels: [], required_labels_warning: nil)
       github_labels = danger.github.pr_labels
 
-      # warn if a PR matches any of the
-      missing_required_labels = required_labels.reject do |required_label|
-        github_labels.any? { |pr_label| pr_label =~ required_label }
-      end
-
-      unless missing_required_labels.empty?
-        missing_labels_str_list = missing_required_labels.map(&:source)
-        warn(required_labels_warning || "PR is missing label(s) matching: #{csv_markdown_list(missing_labels_str_list)}")
-      end
-
-      # A PR shouldn't be merged with the 'DO NOT MERGE' label.
+      # A PR shouldn't be merged with the 'DO NOT MERGE' label
       found_labels = github_labels.select do |github_label|
         do_not_merge_labels.any? { |label| github_label.casecmp?(label) }
       end
 
       failure("This PR is tagged with #{csv_markdown_list(found_labels)} label(s).") unless found_labels.empty?
+
+      # warn if a PR is missing any of the required labels
+      missing_required_labels = required_labels.reject do |required_label|
+        github_labels.any? { |pr_label| pr_label =~ required_label }
+      end
+
+      return if missing_required_labels.empty?
+
+      missing_labels_str_list = missing_required_labels.map(&:source)
+      warn(required_labels_warning || "PR is missing label(s) matching: #{csv_markdown_list(missing_labels_str_list)}")
     end
 
     private
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -2,7 +2,7 @@
 
 module Danger
   # Plugin to check for labels in a PR.
-  class LabelsPRChecker < Plugin
+  class LabelsChecker < Plugin
     DEFAULT_DO_NOT_MERGE_LABEL = 'Do Not Merge'
     DEFAULT_LABELS_WARNING = 'Please add the required labels for this project.'
     DEFAULT_REQUIRED_LABELS = [/.*\S.*/].freeze
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -3,39 +3,47 @@
 module Danger
   # Plugin to check for labels in a PR.
   class LabelsChecker < Plugin
-    DEFAULT_DO_NOT_MERGE_LABEL = 'Do Not Merge'
-    DEFAULT_LABELS_WARNING = 'Please add the required labels for this project.'
-    DEFAULT_REQUIRED_LABELS = [/.*\S.*/].freeze
+    DEFAULT_DO_NOT_MERGE_LABELS = [
+      'Do Not Merge'
+    ]
 
-    # Checks if a PR is missing labels or is marked as 'do not merge'.
+    # Checks if a PR is missing labels or is marked with labels for not merging.
+    # If labels are missing, the plugin will emit a warning. If a label indicating that the PR should not be merged is present,
+    # an error will be emitted, preventing the final PR merge.
     #
-    # @param do_not_merge_label [String] The label indicating that a merge should not be performed.
-    #   Defaults to DEFAULT_DO_NOT_MERGE_LABEL if not provided.
-    # @param required_labels [Array<String>] The list of labels that must be present for a merge operation to proceed.
+    # @param do_not_merge_labels [String] The possible labels indicating that a merge should not be allowed.
+    #   Defaults to DEFAULT_DO_NOT_MERGE_LABELS if not provided.
+    # @param required_labels [Array<String>] The list of RegExes that will prevent the merge if any matching label is present in the PR.
     #   Defaults to an empty array if not provided.
     # @param required_labels_warning [String] The warning message displayed if the required labels are not present.
-    #   Defaults to DEFAULT_LABELS_WARNING if not provided.
+    #   Defaults to showing the provided label regexes.
     #
     # @return [void]
-    def check(do_not_merge_label: DEFAULT_DO_NOT_MERGE_LABEL, required_labels: DEFAULT_REQUIRED_LABELS,
-              required_labels_warning: DEFAULT_LABELS_WARNING)
+    def check(do_not_merge_labels: DEFAULT_DO_NOT_MERGE_LABELS, required_labels: [], required_labels_warning: nil)
       github_labels = danger.github.pr_labels
 
-      pr_has_all_required_labels = github_labels.all? do |pr_label|
-        required_labels.any? do |required_label|
-          pr_label =~ required_label
-        end
+      # warn if a PR matches any of the
+      missing_required_labels = required_labels.reject do |required_label|
+        github_labels.any? { |pr_label| pr_label =~ required_label }
       end
 
-      if github_labels.empty?
-        warn('PR is missing at least one label.')
-      elsif !pr_has_all_required_labels
-        warn(required_labels_warning)
+      unless missing_required_labels.empty?
+        missing_labels_str_list = missing_required_labels.map { |regex| regex.source }
+        warn(required_labels_warning || "PR is missing label(s) matching: #{csv_markdown_list(missing_labels_str_list)}")
       end
 
       # A PR shouldn't be merged with the 'DO NOT MERGE' label.
-      found_label = github_labels.find { |label| do_not_merge_label.casecmp?(label) }
-      failure("This PR is tagged with '#{found_label}' label.") if found_label
+      found_labels = github_labels.select do |github_label|
+        do_not_merge_labels.any? { |label| github_label.casecmp?(label) }
+      end
+
+      failure("This PR is tagged with #{csv_markdown_list(found_labels)} label(s).") unless found_labels.empty?
+    end
+
+    private
+
+    def csv_markdown_list(items)
+      items.map { |item| "`#{item}`" }.join(", ")
     end
   end
 end
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -1,7 +1,20 @@
 # frozen_string_literal: true
 
 module Danger
-  # Plugin to check for labels in a PR.
+  # Plugin for checking labels associated with a pull request.
+  #
+  # @example Checking for specific labels and generating warnings/errors:
+  #   labels_checker.check(
+  #     do_not_merge_labels: ['Do Not Merge'],
+  #     required_labels: ['Bug', 'Enhancement'],
+  #     required_labels_error: 'Please ensure the PR has labels "Bug" or "Enhancement".',
+  #     recommended_labels: ['Documentation'],
+  #     recommended_labels_warning: 'Consider adding the "Documentation" label for better tracking.'
+  #   )
+  #
+  # @see Automattic/dangermattic
+  # @tags github, process
+  #
   class LabelsChecker < Plugin
     DEFAULT_DO_NOT_MERGE_LABELS = [
       'Do Not Merge'
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -5,7 +5,7 @@ module Danger
   class LabelsChecker < Plugin
     DEFAULT_DO_NOT_MERGE_LABELS = [
       'Do Not Merge'
-    ]
+    ].freeze
 
     # Checks if a PR is missing labels or is marked with labels for not merging.
     # If labels are missing, the plugin will emit a warning. If a label indicating that the PR should not be merged is present,
@@ -28,7 +28,7 @@ module Danger
       end
 
       unless missing_required_labels.empty?
-        missing_labels_str_list = missing_required_labels.map { |regex| regex.source }
+        missing_labels_str_list = missing_required_labels.map(&:source)
         warn(required_labels_warning || "PR is missing label(s) matching: #{csv_markdown_list(missing_labels_str_list)}")
       end
 
@@ -43,7 +43,7 @@ module Danger
     private
 
     def csv_markdown_list(items)
-      items.map { |item| "`#{item}`" }.join(", ")
+      items.map { |item| "`#{item}`" }.join(', ')
     end
   end
 end
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -21,31 +21,34 @@ module Danger
     ].freeze
 
     # Checks if a PR is missing labels or is marked with labels for not merging.
-    # If labels are missing, the plugin will emit a warning. If a label indicating that the PR should not be merged is present,
-    # an error will be emitted, preventing the final PR merge.
+    # If recommended labels are missing, the plugin will emit a warning. If a required label is missing, or the PR
+    # has a label indicating that the PR should not be merged, an error will be emitted, preventing the final merge.
     #
     # @param do_not_merge_labels [String] The possible labels indicating that a merge should not be allowed.
     #   Defaults to DEFAULT_DO_NOT_MERGE_LABELS if not provided.
-    # @param required_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that are *required* on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
+    # @param required_labels [Array<Regexp>] The list of Regular Expressions describing all the type of labels that are *required* on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
     #   Defaults to an empty array if not provided.
     # @param required_labels_error [String] The error message displayed if the required labels are not present.
     #   Defaults to a generic message that includes the missing label's regexes.
-    # @param recommended_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that we want a PR to have,
+    # @param recommended_labels [Array<Regexp>] The list of Regular Expressions describing all the type of labels that we want a PR to have,
     # with a warning if it doesn't (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
     #   Defaults to an empty array if not provided.
     # @param recommended_labels_warning [String] The warning message displayed if the recommended labels are not present.
     #   Defaults to a generic message that includes the missing label's regexes.
+    #
+    # @note Tip: if you want to require or recommend "at least one label", you can use
+    #  an array of a single empty regex `[//]` to match "a label with any name".
     #
     # @return [void]
     def check(do_not_merge_labels: DEFAULT_DO_NOT_MERGE_LABELS, required_labels: [], required_labels_error: nil, recommended_labels: [], recommended_labels_warning: nil)
       github_labels = danger.github.pr_labels
 
       # A PR shouldn't be merged with the 'DO NOT MERGE' label
-      found_labels = github_labels.select do |github_label|
+      found_do_not_merge_labels = github_labels.select do |github_label|
         do_not_merge_labels.any? { |label| github_label.casecmp?(label) }
       end
 
-      failure("This PR is tagged with #{markdown_list_string(found_labels)} label(s).") unless found_labels.empty?
+      failure("This PR is tagged with #{markdown_list_string(found_do_not_merge_labels)} label(s).") unless found_do_not_merge_labels.empty?
 
       # fail if a PR is missing any of the required labels
       check_missing_labels(labels: github_labels, expected_labels: required_labels, fail_on_missing: true, custom_message: required_labels_error)
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -16,12 +16,12 @@ module Danger
     # @param required_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that are *required* on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
     #   Defaults to an empty array if not provided.
     # @param required_labels_error [String] The error message displayed if the required labels are not present.
-    #   Defaults to showing the provided label regexes.
+    #   Defaults to a generic message that includes the missing label's regexes.
     # @param recommended_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that we want a PR to have,
     # with a warning if it doesn't (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
     #   Defaults to an empty array if not provided.
     # @param recommended_labels_warning [String] The warning message displayed if the recommended labels are not present.
-    #   Defaults to showing the provided label regexes.
+    #   Defaults to a generic message that includes the missing label's regexes.
     #
     # @return [void]
     def check(do_not_merge_labels: DEFAULT_DO_NOT_MERGE_LABELS, required_labels: [], required_labels_error: nil, recommended_labels: [], recommended_labels_warning: nil)
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -8,6 +8,15 @@ module Danger
     DEFAULT_REQUIRED_LABELS = [/.*\S.*/].freeze
 
     # Checks if a PR is missing labels or is marked as 'do not merge'.
+    #
+    # @param do_not_merge_label [String] The label indicating that a merge should not be performed.
+    #   Defaults to DEFAULT_DO_NOT_MERGE_LABEL if not provided.
+    # @param required_labels [Array<String>] The list of labels that must be present for a merge operation to proceed.
+    #   Defaults to an empty array if not provided.
+    # @param required_labels_warning [String] The warning message displayed if the required labels are not present.
+    #   Defaults to DEFAULT_LABELS_WARNING if not provided.
+    #
+    # @return [void]
     def check(do_not_merge_label: DEFAULT_DO_NOT_MERGE_LABEL, required_labels: DEFAULT_REQUIRED_LABELS,
               required_labels_warning: DEFAULT_LABELS_WARNING)
       github_labels = danger.github.pr_labels
--- a/lib/dangermattic/plugins/labels_checker.rb
+++ b/lib/dangermattic/plugins/labels_checker.rb
@@ -13,13 +13,18 @@ module Danger
     #
     # @param do_not_merge_labels [String] The possible labels indicating that a merge should not be allowed.
     #   Defaults to DEFAULT_DO_NOT_MERGE_LABELS if not provided.
-    # @param required_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that we want to be required on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
+    # @param required_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that are *required* on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
     #   Defaults to an empty array if not provided.
-    # @param required_labels_warning [String] The warning message displayed if the required labels are not present.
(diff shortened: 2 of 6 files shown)


lib/danger_plugin.rb
```diff
@@ -1,3 +1,4 @@
 # frozen_string_literal: true
 
+require 'dangermattic/plugins/labels_pr_checker'
 require 'dangermattic/plugins/view_changes_need_screenshots'
```
AliSoftware: I wonder if at some point we shouldn't just write `require 'dangermattic/plugins/*` to automatically require all our plugins? (assuming that `*` works in `require`) 🤔

lib/dangermattic/plugins/labels_pr_checker.rb
```diff
@@ -0,0 +1,24 @@
...
+      # A PR shouldn't be merged with the 'DO NOT MERGE' label.
+      do_not_merge_label = github_labels.find { |label| do_not_merge_label?(label) }
+      failure("This PR is tagged with '#{do_not_merge_label}' label.") if do_not_merge_label
+    end
+
+    private
+
+    def do_not_merge_label?(label)
+      lc_label = label.downcase
+      lc_label.include?('do not merge') || lc_label.include?('not ready for merge')
```
AliSoftware: I'd suggest to create a `DO_NOT_MERGE_LABELS = ['Do Not Merge', 'Not Ready for Merge'].freeze` constant, then test using something like:
```ruby
def do_not_merge_label?(label)
  DO_NOT_MERGE_LABELS.any? { |l| label.casecmp?(l) }
end
```

lib/dangermattic/plugins/labels_pr_checker.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin to check for labels in a PR.
+  class LabelsPRChecker < Plugin
+    # Checks if a PR is missing labels or is marked as 'do not merge'.
+    def check_labels
+      github_labels = danger.github.pr_labels
+
+      warn('PR is missing at least one label.') if github.pr_labels.empty?
```
AliSoftware: Might be for a subsequent PR—when you'll consider refactoring some existing rules as opposed to just porting the existing Peril ones _as is_ like here—but it could be worth merging this Danger rule with the future port of https://github.com/Automattic/peril-settings/blob/master/org/issue/label-woo.ts and https://github.com/Automattic/peril-settings/blob/master/org/issue/label.ts to make a single, more flexible/generic rule instead.

For example, in WCiOS, we have a set of GitHub Labels prefixed by `feature:`, and also a set of labels prefixed with `type:`, and we require all issues to have at least one of those `feature: *` labels and one of those `type: *` labels.

Maybe your dangermattic Plugin here could factor all that into one single plugin that could be applied to both PRs and Issues, and configurable for each repo's labels, instead of being specific for PRs with the hardcoded value of Do Not Merge?

I'm thinking something like this for the call site in the client's Dangerfile:

```ruby
labels_checker.do_not_merge_label = 'Do Not Merge'
labels_checker.require_label(/feature: .*/)
labels_checker.require_label(/type: .*/)
labels_checker.check()
```

Or something like:
```ruby
labels_checker.check(
  do_not_merge_label: 'Do Not Merge',
  required_labels: [/feature: .*/, /type: .*/]
)
```

spec/labels_pr_checker_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        @plugin = @dangerfile.labels_pr_checker
+      end
+
+      it 'returns a warning when a PR does not have labels' do
+        pr_labels = []
+        allow(@plugin.github).to receive(:pr_labels).and_return(pr_labels)
+
+        @plugin.check_labels
+
+        expect(@dangerfile.status_report[:warnings].count).to eq 1
```
AliSoftware: Similar to my comment on your other PR, I'd suggest your `expect` to check the actual content of the status_report, and especially the warning messages they contain, rather than assert on their count only.

lib/danger_plugin.rb
iangmaia: ~Hmm yeah, some quick searches show that perhaps `require_all 'dangermattic/plugins/*.rb'` could work 🤔 gonna give it a try 👍~ 
Update: `require_all` is a gem. `Dir[File.join(__dir__, 'dangermattic/plugins', '*.rb')].each { |file| require file }` works.

lib/dangermattic/plugins/labels_pr_checker.rb
iangmaia: > your dangermattic Plugin here could factor all that into one single plugin that could be applied to both PRs and Issues, and configurable for each repo's labels, instead of being specific for PRs with the hardcoded value of Do Not Merge

That's exactly what I think as well -- some of the previous Peril rules are perhaps too tied to a specific process or even specific app practices. We should try to generalise them to make them applicable at least to all A8c apps.

For now this was a 1:1 port, but I think your suggestion makes a lot of sense and shouldn't be a lot of trouble to generalise from the beginning in this case. Let me give it a try!

iangmaia: I've changed the plugin to generalise it a bit more and receive these parameters.

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,41 @@
...
+    #
+    # @param do_not_merge_label [String] The label indicating that a merge should not be performed.
+    #   Defaults to DEFAULT_DO_NOT_MERGE_LABEL if not provided.
+    # @param required_labels [Array<String>] The list of labels that must be present for a merge operation to proceed.
+    #   Defaults to an empty array if not provided.
+    # @param required_labels_warning [String] The warning message displayed if the required labels are not present.
+    #   Defaults to DEFAULT_LABELS_WARNING if not provided.
+    #
+    # @return [void]
+    def check(do_not_merge_label: DEFAULT_DO_NOT_MERGE_LABEL, required_labels: DEFAULT_REQUIRED_LABELS,
```
AliSoftware: 💭  I think it could be benefit from allowing potentially multiple `do_not_merge_labels` instead of just one. Not really because repos might have multiple labels explicitly named something like "Do Not Merge" with variations, no, but more because there might be other labels for which we might want Danger to block a PR merge.

For example, imagine a repo where we'd have GitHub labels like "Do Not Merge" but also "Requires Design Review" and "Pending Product Team's Go/NoGo" or "Depends on Other PR" and similar, that would then make sense to be able to pass all of those labels to our Plugin so it blocks the merge if any of those is present 🙂

AliSoftware: 💭 🔜  As a side note, idea for future Plugins while I think about it: it would be nice if we could standardize, if possible and if it makes sense, how we call the main entry point of all our dangermattic plugins. Like call them all `check(…)` or `run(…)` or similar, just for consistency.

Obviously, this is just a thought for later — especially only once you have a couple of Plugins merged into `trunk` and we can think of a good name that would work for all of them (maybe `check()` will actually be the right one); for example maybe you could keep that in mind for when you'll also work on adding YARD/documentation (calls to `yard doc` / `yard stats --list-undoc`) from CI (to check and generate the plugins' documentation) later in your project's plan?

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin to check for labels in a PR.
+  class LabelsChecker < Plugin
+    DEFAULT_DO_NOT_MERGE_LABEL = 'Do Not Merge'
+    DEFAULT_LABELS_WARNING = 'Please add the required labels for this project.'
```
AliSoftware:
```suggestion
    DEFAULT_LABELS_WARNING = 'Some required GitHub labels are missing, please add them as appropriate.'
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,41 @@
...
+    DEFAULT_REQUIRED_LABELS = [/.*\S.*/].freeze
```
AliSoftware: I'm not sure I'm following why this would make sense as a default set of required label.
 - If I understand this regex correctly, it's gonna match zero-or-more characters (`.*`) followed by a non-whitespace character (`\S`) , followed by zero-or-more characters again. Which basically matches everything (except labels whose name is composed entirely of whitespaces, which is unlikely and probably not allowed by GitHub anyway)
 - Besides, at first I thought the goal of using a "match-all" regex as default would be a derived way to check if "PR has at least one label", which I think should work and could be a good solution, except that:
    - In practice that's not how you use this, since you hardcoded a `if github_labels.empty?` with `warn('PR is missing at least one label.')` below instead
    - And a simpler Regex like `/.+/` should be enough to do that trick anyway I think

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,41 @@
...
+      github_labels = danger.github.pr_labels
+
+      pr_has_all_required_labels = github_labels.all? do |pr_label|
+        required_labels.any? do |required_label|
+          pr_label =~ required_label
+        end
+      end
+
+      if github_labels.empty?
+        warn('PR is missing at least one label.')
```
AliSoftware: Do we want that requirement to be hardcoded? As in, what if a client repo wanted to use that Plugin only so that it would block a PR from being merged if it has one of the `do_not_merge_labels`, but without really needing to enforce all PRs to have at least one label otherwise?

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,41 @@
...
+    # @return [void]
+    def check(do_not_merge_label: DEFAULT_DO_NOT_MERGE_LABEL, required_labels: DEFAULT_REQUIRED_LABELS,
+              required_labels_warning: DEFAULT_LABELS_WARNING)
+      github_labels = danger.github.pr_labels
+
+      pr_has_all_required_labels = github_labels.all? do |pr_label|
+        required_labels.any? do |required_label|
+          pr_label =~ required_label
+        end
+      end
```
AliSoftware: Would be nice to capture the list of missing required labels, so we can print it in the error message
```suggestion
      missing_required_labels = required_labels.reject do |required_label|
        github_labels.any? { |pr_label| pr_label =~ required_label }
      end
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      if github_labels.empty?
+        warn('PR is missing at least one label.')
+      elsif !pr_has_all_required_labels
+        warn(required_labels_warning)
```
AliSoftware: Would be nice to print the list of missing required labels, so that the PR author knows what's missing and so that the warning is more actionable, otherwise I think it would feel pretty obscure?
```ruby
unless missing_required_labels.empty?
  warn("PR is missing label(s) matching: #{missing_required_labels})
end
```

lib/dangermattic/plugins/labels_checker.rb
AliSoftware: Also, it would be nice for the Plugin to support the elements of `required_labels` to be a either a `RegExp` or just a plain `String` (and in such case, we'd check for an exact match instead of `~=`). Probably a rare use case to require a specific and fixed label on all PRs tho 🤔

AliSoftware: Another use case I just thought of… but that can probably wait for the second iteration of this Plugin, is to conditionalize the required labels based on the base branch.

For example, if the PR targets the branch `release/*` (which corresponds to the code-freeze in most of our repos, similar to `beta` branch in Tumblr), then require either the `bug` or `bugfix-exemption` label, otherwise not. Maybe using an API like this?
```ruby
labels_checker.check(required_labels: {
  /^release\/.*/: /^(bug|bugfix-exemption)$/,
  'develop': /^feature: .*/
})
```
Though maybe it would be simpler to let the call site / client repo do the test in their `Dangerfile` themselves, like this?
```ruby
if github.branch_for_base.start_with?('release/')
  labels_checker.check(required_labels: [/^(bug|bugfix-exemption)$/])
else
  labels_checker.check(required_labels: [/^feature:/])
end
```
Probably a matter of balance to find between how much we want to be within the Plugin (to avoid each client repo to repeat the same kind of code pattern over and over) vs how much we want to be written in the `Dangerfile` (which could allow more flexibility, but also might lead to code repetition between `Dangerfiles` and kinda defeats part of the benefits of DRY-ing code into a Plugin)

I guess we'll see as we start adopting this plugin in repos how we should make that plugin evolve, tho, so maybe better to wait and see concrete use cases first?

lib/dangermattic/plugins/labels_checker.rb
AliSoftware: Or maybe even get rid of that constant, keep the default value to `nil` in the method signature (as in: `check(…, required_labels: nil)`), and generate a _dynamic_ message like `"PR is missing label(s) matching: #{missing_required_labels}"` if no custom message is provided?

lib/dangermattic/plugins/labels_checker.rb
iangmaia: I thought about that, but showing regexes could be a bit cryptic as well 🤔 (well, it is still better than nothing). But that's the main reason I introduced the `required_labels_warning` parameter as well, so that the expected labels could be set inside a free text message from outside of the plugin, in addition to just providing regexes for validation.

lib/dangermattic/plugins/labels_checker.rb
iangmaia: >  imagine a repo where we'd have GitHub labels like "Do Not Merge" but also "Requires Design Review" and "Pending Product Team's Go/NoGo" or "Depends on Other PR" and similar, that would then make sense to be able to pass all of those labels to our Plugin so it blocks the merge

Ah yes, that's a nice idea! In such cases one could just add a 'Do Not Merge', but the actual intention will be lost.

> it would be nice if we could standardize [...] how we call the main entry point of all our dangermattic plugins.

> especially only once you have a couple of Plugins merged into trunk and we can think of a good name that would work for all of them

I agree -- I think we'll have a better view of everything once we have a bunch of plugins together on trunk. I'll keep that in mind, thanks!

lib/dangermattic/plugins/labels_checker.rb
AliSoftware: Yeah, [hence my other suggestion above](https://github.com/Automattic/dangermattic/pull/7#discussion_r1262747487) which would allow the default message to print the expected regexes if no explicit message was provided, while still letting the called provide an explicit warning message to replace it if they prefer.

Then this part of the code would look like:
```ruby
unless missing_required_labels.empty?
  warn(required_labels_warning || "PR is missing label(s) matching: #{missing_required_labels})
end
```

lib/dangermattic/plugins/labels_checker.rb
iangmaia: Ah, nice suggestion! I'll do exactly that: `nil` default and generate the message if none is used. Thanks!

lib/dangermattic/plugins/labels_checker.rb
iangmaia: Yes, I [just seen it now](https://github.com/Automattic/dangermattic/pull/7#discussion_r1262792975)! 🙂

lib/dangermattic/plugins/labels_checker.rb
AliSoftware: I think the simpler approach here would be to set `DEFAULT_REQUIRED_LABELS = [/.*/].freeze # require at least one label, without any constraint on which one`, or maybe make `[]` be the default instead? And then remove the special case of `github_labels.empty?` check on line 30, to instead let the more generic check on the required_label to take care of that case.

 - If one wants to use the plugin just for the do_not_merge labels, they can use `[]` for the required labels
 - If one wants to use the plugin for requiring at least one label, without any rule about which label that is, they can use `[/.*/]`
 - If one wants to request a specific set of labels, then one could use `[/^feature:/, /^type:/]`, and then the case of "require at least one label overall" would already be included in the case of "require at least one feature label and one type label" anyway

lib/dangermattic/plugins/labels_checker.rb
iangmaia: > Probably a matter of balance to find between how much we want to be within the Plugin (to avoid each client repo to repeat the same kind of code pattern over and over) vs how much we want to be written in the Dangerfile

Exactly. My intuition for this specific case is that the plugins should fulfil the general case as much as possible, with this kind of check done outside of the plugin and avoiding conditional logic based on external factors -- the conditional check could also be part of a common `Dangerfile` that repos can include, if it applies to enough applications.

But as you said, I think we need to start seeing once we start adopting it in the repos to have a better idea.

lib/dangermattic/plugins/labels_checker.rb
iangmaia: > And then remove the special case of github_labels.empty? check on line 30, to instead let the more generic check on the required_label to take care of that case.

I think that's a better and more general solution (also addressing your comment [here](https://github.com/Automattic/dangermattic/pull/7#discussion_r1262727924)). Having at least one label was the hard requirement on the Peril rule, but I think it is nicer (and a better strategy in general) to make it more flexible if we can so also more repos can adopt these checks. I liked your suggestion, it's a good balance of flexibility and conventions 👍

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,49 @@
...
+      'Do Not Merge'
+    ].freeze
+
+    # Checks if a PR is missing labels or is marked with labels for not merging.
+    # If labels are missing, the plugin will emit a warning. If a label indicating that the PR should not be merged is present,
+    # an error will be emitted, preventing the final PR merge.
+    #
+    # @param do_not_merge_labels [String] The possible labels indicating that a merge should not be allowed.
+    #   Defaults to DEFAULT_DO_NOT_MERGE_LABELS if not provided.
+    # @param required_labels [Array<String>] The list of RegExes that will prevent the merge if any matching label is present in the PR.
```
AliSoftware:
```suggestion
    # @param required_labels [Array<RexExp>] The list of RegExes that will prevent the merge if any matching label is present in the PR.
```

AliSoftware:
```suggestion
    # @param required_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that we want to be required on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      # warn if a PR is missing any of the required labels
+      missing_required_labels = required_labels.reject do |required_label|
+        github_labels.any? { |pr_label| pr_label =~ required_label }
+      end
+
+      return if missing_required_labels.empty?
+
+      missing_labels_str_list = missing_required_labels.map(&:source)
+      warn(required_labels_warning || "PR is missing label(s) matching: #{csv_markdown_list(missing_labels_str_list)}")
```
AliSoftware: Since the API calls them `required_labels` I'd expect those to actually be **_required_**, i.e. to make the PR fail with a `failure`, not a `warn`.

I can imagine some repos only wanting this as a warning though to be less strict. So maybe that should be a parameter in the API, like `fail_on_missing_labels: true`? Or alternatively, provide both `required_labels: []` and `recommended_labels: []` parameters, with `required_labels` triggering `failure` if they're missing while `recommended_labels` only triggering `warn` (that way the callers can mix both cases if they need to, making the plugin all more flexible).

wdyt?

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    private
+
+    def csv_markdown_list(items)
```
AliSoftware: Nitpick: this method name initially made me think that you had some CSV file format involved in there. Maybe just `def markdown_list(items)` would be enough while avoiding the risk of confusion? 🤷

lib/dangermattic/plugins/labels_checker.rb
iangmaia: That's an interesting point. I do wonder how often it happens that we want a label to cause a failure (and potentially even preventing a merge) -- it feels too strict of a process, but of course in the end this could be up to the host projects to decide.

As I don't think this is happening now in our repos, perhaps this might be a case where we need to see how the usage is going to be and, if it makes sense (or if it in fact improves a process), we can add it?
It's not even like it's a lot more work, of course, but I think it would be good to keep the API simple to start.

AliSoftware: Here's [a case where missing a required label makes the check fail, not warn](https://github.com/Automattic/peril-settings/blob/master/org/issue/label.ts#L13).

_Granted, this check is only applied on issues, not PRs, but that's the same idea. In this case, I _think_ this was enforced because the QE team uses (used?) GitHub labels to classify the bugs reported in the app, and create stats and priorities based on those, hence them being strongly required here I guess._

iangmaia: I see a report [here](https://github.com/wordpress-mobile/WordPress-Android/issues/18767#issuecomment-1633490431)! Though not sure what will be the implication of fail compared to a warning for issues (guessing no difference)?

I'm gonna give the `recommended_labels` approach a try, it shouldn't complicate the API that much.
But that also makes me think that, if we also have the option to fail on certain label patterns, wouldn't that make the `do_not_merge_labels` parameter obsolete?

AliSoftware: Well the `do_not_merge_labels` work in the opposite way: if the label is set, then we `fail`. As opposed to `required_labels` which fail only if the label is _not_ set. So no, I don't think that wouldn't make it obsolete.

iangmaia: Ah, of course, I got confused for a moment 🙃 Both the label regex arrays parameters would be checking for labels we WANT the PR to have, otherwise failing/warning; while the "do not merge" labels are for potentially preventing a PR to be merged, always failing if they are present.

iangmaia: Committed some changes [here](https://github.com/Automattic/dangermattic/pull/7/commits/b1e600005b556ce481c4fdd7bed4d0415349d281) adding the new parameters.

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,67 @@
...
+    # Checks if a PR is missing labels or is marked with labels for not merging.
+    # If labels are missing, the plugin will emit a warning. If a label indicating that the PR should not be merged is present,
+    # an error will be emitted, preventing the final PR merge.
+    #
+    # @param do_not_merge_labels [String] The possible labels indicating that a merge should not be allowed.
+    #   Defaults to DEFAULT_DO_NOT_MERGE_LABELS if not provided.
+    # @param required_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that are *required* on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
+    #   Defaults to an empty array if not provided.
+    # @param required_labels_error [String] The error message displayed if the required labels are not present.
+    #   Defaults to showing the provided label regexes.
```
AliSoftware:
```suggestion
    #   Defaults to a generic message that includes the missing label's regexes.
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,67 @@
...
+    # @param recommended_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that we want a PR to have,
+    # with a warning if it doesn't (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
+    #   Defaults to an empty array if not provided.
+    # @param recommended_labels_warning [String] The warning message displayed if the recommended labels are not present.
+    #   Defaults to showing the provided label regexes.
```
AliSoftware:
```suggestion
    #   Defaults to a generic message that includes the missing label's regexes.
```

spec/labels_checker_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+
+          expect(@dangerfile.status_report[:warnings]).to be_empty
+          expect(@dangerfile.status_report[:errors]).to eq([custom_labels_error])
+        end
+
+        it 'does nothing when custom required labels are correctly set in the PR' do
+          pr_labels = [
+            'feature: time travel',
+            'type: prototype'
+          ]
```
AliSoftware: Just so that this test also ensures that our implementation does not accidentally require **_all_** labels to meet the requirements, but only that we have at least one label meeting each requirement (in other words, that the behavior is not "all labels need to meet the requirements", but "each requirement needs to be met by at least one label")

```suggestion
          pr_labels = [
            'feature: time travel',
            'type: prototype',
            'some other label'
          ]
```

spec/labels_checker_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+          expect(@dangerfile.status_report[:warnings]).to be_empty
+        end
+
+        it 'returns an error containing the required labels when a PR does not meet the label requirements' do
+          pr_label = 'random other label'
+          allow(@plugin.github).to receive(:pr_labels).and_return([pr_label])
+
+          @plugin.check(
+            required_labels: [/feature: .*/]
+          )
```
AliSoftware: I'd suggest to also test test the case of only some of the labels meeting the requirements but not all.

In fact, none of your test checks the case for multiple `required_labels`, while we need to make sure that:
 - it's not interpreted as "all labels must match one of those RegExes" (while we want to allow `'random other label'` even if it doesn't match, as long as we _also_ have one label like `feature: magic` too that matches)
 - nor being interpreted as "we need one label that matches any of those RegExes" (i.e. if we have `[/feature: .*/, /type: .*/]` then we want a label of each, and just having one label that matches one of the two should not be enough
 - but instead properly interpreted as "we require at least one label matching for each RegEx we provide"

In other words, we need test cases with `required_labels: [/^feature:/, /^type:/]` that would:
 - Fail if `pr_labels = ['random label', 'feature: magic', 'wizard needed']` (ensuring we detect that we're missing a label with `type: .*`)
 - Fail if `pr_labels = ['feature: magic']` (ensuring that we don't accidentally interpret the code as "all labels must match at least one of the `required_labels`", but quite the opposite of "all `required_labels` must match at least one label")
 - Pass if `pr_labels = ['random 1', 'feature: magic', 'random 2', 'type: fantasy', 'wizard needed']`, to ensure that not only we check that both `/^feature:/` and `/^type:/` have a match, but also that we're ok if the other labels don't match, as long as one label matched each

spec/labels_checker_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+        @dangerfile = testing_dangerfile
+        @plugin = @dangerfile.labels_checker
+      end
+
+      context 'with required labels' do
+        it 'returns a custom error when a PR does not have at least one label' do
+          allow(@plugin.github).to receive(:pr_labels).and_return([])
+
+          error = 'PR is missing at least one label.'
+          @plugin.check(required_labels: [/.*/], required_labels_error: error)
```
AliSoftware: Note that the RegExp are not anchored in Ruby, so they can match just part of the string instead of the whole string. Which that means that `/x/.match?('abcxdef')` will match too.

As a result:
 - We should use `^` and `$` to anchor the regex if we want to make sure that `/feature: .*/` won't risk matching a label like `'secret feature: admin mode'`.
 - Or if the goal is just to match labels that starts with a specific string, we don't need to use `.*$` to match up to the end, and we can just use `/^feature:/` for the RegExp
 - And similarly here, using `//` as an empty RegExp should match _any_ string (i.e. `//.match?(text)` will always return true), so you could use that instead of `/.*/`.

iangmaia: Nice observation 👍

spec/labels_checker_spec.rb
iangmaia: Thanks a lot! I added more data to cover these scenarios: f717b268ebeea34e6b9409708eea13b801dd3b4b

lib/danger_plugin.rb
```diff
@@ -1,6 +1,3 @@
 # frozen_string_literal: true
 
-require 'dangermattic/plugins/manifest_pr_checker'
-require 'dangermattic/plugins/milestone_checker'
-require 'dangermattic/plugins/pr_size_checker'
-require 'dangermattic/plugins/view_changes_need_screenshots'
+Dir[File.join(__dir__, 'dangermattic/plugins', '*.rb')].each { |file| require file }
```
AliSoftware, line 3: You could also make use of `require_relative` instead:
```suggestion
Dir['dangermattic/plugins/*.rb'].each do |file|
  require_relative file
end
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,80 @@
...
+  # @tags github, process
+  #
+  class LabelsChecker < Plugin
+    DEFAULT_DO_NOT_MERGE_LABELS = [
+      'Do Not Merge'
+    ].freeze
+
+    # Checks if a PR is missing labels or is marked with labels for not merging.
+    # If labels are missing, the plugin will emit a warning. If a label indicating that the PR should not be merged is present,
+    # an error will be emitted, preventing the final PR merge.
```
AliSoftware:
```suggestion
    # If recommended labels are missing, the plugin will emit a warning. If a required label is missing, or the PR
    # has a label indicating that the PR should not be merged, an error will be emitted, preventing the final merge.
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,80 @@
...
+    #
+    # @param do_not_merge_labels [String] The possible labels indicating that a merge should not be allowed.
+    #   Defaults to DEFAULT_DO_NOT_MERGE_LABELS if not provided.
+    # @param required_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that are *required* on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
```
AliSoftware: Ruby's regular expression type is named `Regexp` not `RegExp` (I get tripped by this all the time 😅 )
```suggestion
    # @param required_labels [Array<Regexp>] The list of Regular Expressions describing all the type of labels that are *required* on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,80 @@
...
+    #   Defaults to an empty array if not provided.
+    # @param required_labels_error [String] The error message displayed if the required labels are not present.
+    #   Defaults to a generic message that includes the missing label's regexes.
+    # @param recommended_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that we want a PR to have,
```
AliSoftware:
```suggestion
    # @param recommended_labels [Array<Regexp>] The list of Regular Expressions describing all the type of labels that we want a PR to have,
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,80 @@
...
+    # @return [void]
+    def check(do_not_merge_labels: DEFAULT_DO_NOT_MERGE_LABELS, required_labels: [], required_labels_error: nil, recommended_labels: [], recommended_labels_warning: nil)
+      github_labels = danger.github.pr_labels
+
+      # A PR shouldn't be merged with the 'DO NOT MERGE' label
+      found_labels = github_labels.select do |github_label|
+        do_not_merge_labels.any? { |label| github_label.casecmp?(label) }
+      end
+
+      failure("This PR is tagged with #{markdown_list_string(found_labels)} label(s).") unless found_labels.empty?
```
AliSoftware: Now that we extended this check with the concept of required and recommended labels, maybe find a more suited name for this variable

```suggestion
      found_do_not_merge_labels = github_labels.select do |github_label|
        do_not_merge_labels.any? { |label| github_label.casecmp?(label) }
      end

      failure("This PR is tagged with #{markdown_list_string(found_labels)} label(s).") unless found_do_not_merge_labels.empty?
```

lib/dangermattic/plugins/labels_checker.rb
```diff
@@ -0,0 +1,80 @@
...
+    # @param required_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that are *required* on PR (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
+    #   Defaults to an empty array if not provided.
+    # @param required_labels_error [String] The error message displayed if the required labels are not present.
+    #   Defaults to a generic message that includes the missing label's regexes.
+    # @param recommended_labels [Array<RegExp>] The list of Regular Expressions describing all the type of labels that we want a PR to have,
+    # with a warning if it doesn't (e.g. `[/^feature:/, `/^type:/]` or `bug|bugfix-exemption`).
+    #   Defaults to an empty array if not provided.
+    # @param recommended_labels_warning [String] The warning message displayed if the recommended labels are not present.
+    #   Defaults to a generic message that includes the missing label's regexes.
+    #
```
AliSoftware, line 38:
```suggestion
    #
    # @note Tip: if you want to require or recommend "at least one label", you can use
    #  an array of a single empty regex `[//]` to match "a label with any name".
    #
```

AliSoftware (approved):
LGTM, only left some rewording suggestions but otherwise approving 👍

lib/danger_plugin.rb
iangmaia: This is something I tried in the early versions but it didn't work -- just tried here again and the tests fail with `uninitialized constant Danger::<plugin>`.

lib/dangermattic/plugins/labels_checker.rb
iangmaia: Oof, true.

PR #5, opened 2023-06-30
Make RuboCop setup stricter

mokagio:
There are eleven violations left to address, but I wanted to get some thoughts on this PR before moving forward with them.

@iangmaia I noticed Rake runs RuboCop only on lib and spec. I'm curious about your rationale for this.

Personally, I think we should run it on every file we can, and that's the assumption on which I based these changes.

### Remaining violations

```
Inspecting 8 files
..CW....

Offenses:

Rakefile:11:1: C: Rake/Desc: Describe the task with desc method.
task :spec do
^^^^^^^^^^
dangermattic.gemspec:1:1: W: Gemspec/RequiredRubyVersion: required_ruby_version should be specified.
dangermattic.gemspec:32:3: C: Gemspec/DevelopmentDependencies: Specify development dependencies in Gemfile.
  spec.add_development_dependency 'bundler', '~> 2.0'
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
dangermattic.gemspec:33:3: C: Gemspec/DevelopmentDependencies: Specify development dependencies in Gemfile.
  spec.add_development_dependency 'rake', '~> 13.0'
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
dangermattic.gemspec:36:3: C: Gemspec/DevelopmentDependencies: Specify development dependencies in Gemfile.
  spec.add_development_dependency 'rspec', '~> 3.4'
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
dangermattic.gemspec:40:3: C: Gemspec/DevelopmentDependencies: Specify development dependencies in Gemfile.
  spec.add_development_dependency 'rubocop-rake'
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
dangermattic.gemspec:41:3: C: Gemspec/DevelopmentDependencies: Specify development dependencies in Gemfile.
  spec.add_development_dependency 'rubocop-rspec'
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
dangermattic.gemspec:42:3: C: Gemspec/DevelopmentDependencies: Specify development dependencies in Gemfile.
  spec.add_development_dependency 'yard'
[...]

PR #8, opened 2023-07-13
Improve RuboCop Setup

iangmaia:
This PR is an iteration over @mokagio's PR: https://github.com/Automattic/dangermattic/pull/5.

I've brought over some RuboCop rules configs from [release-toolkit](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/.rubocop.yml) that seemed sensible and fixed a few RuboCop offences in what was already merged to `trunk`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,7 @@
+require:
+  - rubocop-rake
+  - rubocop-rspec
+
+
+AllCops:
+  NewCops: enable
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,52 @@ require:
   - rubocop-rake
   - rubocop-rspec
 
-
 AllCops:
   NewCops: enable
+
+Gemspec/DevelopmentDependencies:
+  EnforcedStyle: gemspec
+
+########## Metrics / Max Lengths Rules
+
+Layout/LineLength:
+  Max: 300
+
+Metrics/AbcSize:
+  Max: 140
+
+Metrics/BlockLength:
+  Max: 80
+  Exclude:
+   - spec/**/*_spec.rb
+   - spec/**/shared_examples_*.rb
+
+Metrics/ClassLength:
+  Max: 300
+
+Metrics/MethodLength:
+  Max: 150
+
+Metrics/ModuleLength:
+  Max: 300
+
+Metrics/ParameterLists:
+  Max: 10
+
+########## Metrics / Complexity
+
+Metrics/CyclomaticComplexity:
+  Max: 20
+
+Metrics/PerceivedComplexity:
+  Max: 20
+
+########## RSpec Rules
+
+# We are not strict to the point that we want to enforce super-short `it '…' do … end` blocks in our specs.
+RSpec/ExampleLength:
+  Enabled: false
+
+# Same for number of let/subject
+RSpec/MultipleMemoizedHelpers:
+  Enabled: false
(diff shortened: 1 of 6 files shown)


dangermattic.gemspec
```diff
@@ -53,4 +54,5 @@ Gem::Specification.new do |spec|
   #
   # This will stop test execution and let you inspect the results
   spec.add_development_dependency 'pry'
+  spec.metadata['rubygems_mfa_required'] = 'true'
```
AliSoftware: I thought we removed that one from release-toolkit (which you presumably copied that from?), because that would prevent us from using CI to release the gem to rubygems on new tags automatically?

spec/view_changes_need_screenshots_spec.rb
```diff
@@ -49,60 +49,60 @@ module Danger
         @plugin = @dangerfile.view_changes_need_screenshots
       end
 
-      context 'iOS PR has view code changes in Swift files' do
+      context 'with an iOS PR having view code changes in Swift files' do
```
AliSoftware: This change and all the similar ones below about the convention of beginning `context` names using `when` or `with` kinda self-advocate for breaking down those contexts in 2 levels:
```ruby
context 'when checking an iOS PR' do
  context 'with View code changes in Swift files' do … end
  context 'with Button code changes in Swift files' do … end
  context 'with View code changes in ObjC files' do … end
  context 'with Button code changes in Swift files' do … end
end

context 'when checking an Android PR' do
  context 'with View code changes in Kotlin files' do … end
  context 'with Button code changes in Kotlin files' do … end
  context 'with View code changes in Java files' do … end
  context 'with Button code changes in Java files' do … end
end
```

dangermattic.gemspec
AliSoftware: See also https://github.com/wordpress-mobile/release-toolkit/pull/475

iangmaia: Oh, this came from Gio's [original PR](https://github.com/Automattic/dangermattic/pull/5/files#diff-72eca2967c1407b19db3f71a0d5199513786a05fcdd2f1f2d9e4b0e782ba72cfR56). I understood it was a RuboCop offence, but let me try removing it and see how it goes.

iangmaia: Indeed, it is reported by RuboCop and auto-correctable:
```
dangermattic.gemspec:9:1: W: [Correctable] Gemspec/RequireMFA: metadata['rubygems_mfa_required'] must be set to 'true'.
```

AliSoftware: We just have to disable the `Gemspec/RequireMFA` rule in `.rubocop.yml` then (maybe with a `# comment` as to why next to it).

spec/view_changes_need_screenshots_spec.rb
iangmaia: Nice one 👍

dangermattic.gemspec
AliSoftware: You can copy/paste the comment from https://github.com/wordpress-mobile/release-toolkit/pull/464/files#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5cR61-R65 if you need inspiration 😉

.rubocop.yml
```diff
@@ -0,0 +1,59 @@
+require:
+  - rubocop-rake
+  - rubocop-rspec
+
+AllCops:
+  NewCops: enable
+
+########## Gemspec Rules
+
+# Disabling this as all the gem publishing will be done within CI and doesn't allow for user input such as an MFA code.
```
AliSoftware, line 10: 👍

AliSoftware (approved):

PR #9, opened 2023-07-20
Plugin for iOS and Android Release Checks

iangmaia:
This PR ports the rules related to releases that are currently implemented in the following Peril rules:
- https://github.com/Automattic/peril-settings/blob/master/org/pr/ios-macos.ts
- https://github.com/Automattic/peril-settings/blob/master/org/pr/android.ts
- https://github.com/Automattic/peril-settings/blob/master/org/pr/release-notes.ts

I tried to make them more generic along the way as some of them seem closely related to the conventions and general process for WordPress and WooCommerce.
There is a check for `Releases` in some of these Peril rules, which is a logic to prevent the checks to run. My idea is to have these checks performed either in the host project's `Dangerfile` or in a common `Dangerfile` provided by this repo that can be included in the host project's `Dangerfile`, in such a way that the plugins won't need more specific checks as this one.

This PR introduces the following plugins:
- `CommonReleaseChecks`: contains generic checks, in a process agnostic way so that it can be used independently.
- `AndroidReleaseCheck` and `IosReleaseCheck`: they both use `CommonReleaseChecks` to perform iOS/Android specific checks

The remainder rules from the files mentioned above will follow in another PR.

### DIFF
--- a/lib/dangermattic/plugins/common/common_release_checks.rb
+++ b/lib/dangermattic/plugins/common/common_release_checks.rb
@@ -14,7 +14,7 @@ module Danger
   # @example Checking if release notes and store strings have changed:
   #          common_release_checks.check_release_notes_and_store_strings(
   #            release_notes_file: 'metadata/release_notes.txt',
-  #            store_strings_file: 'metadata/PlayStoreStrings.po'
+  #            po_file: 'metadata/PlayStoreStrings.po'
   #          )
   #
   # @example Checking for changes in internal release notes:
@@ -69,20 +69,20 @@ module Danger
     # @param release_notes_file [String] The name of the release notes file that should be checked for modifications.
     #   Example: 'metadata/release_notes.txt'
     #
-    # @param store_strings_file [String] The name of the store strings file that should be checked for modifications.
+    # @param po_file [String] The name of the store strings file that should be checked for modifications.
     #   Example: 'metadata/PlayStoreStrings.po'
     #
     # @example Check if the release notes file 'release_notes.txt' is modified, and the 'PlayStoreStrings.po' file is not modified, posting a message:
-    #          check_release_notes_and_store_strings(release_notes_file: 'release_notes.txt', store_strings_file: 'PlayStoreStrings.po')
+    #          check_release_notes_and_store_strings(release_notes_file: 'release_notes.txt', po_file: 'PlayStoreStrings.po')
     #
     # @return [void]
-    def check_release_notes_and_store_strings(release_notes_file:, store_strings_file:)
-      has_modified_release_notes = danger.git.modified_files.any? { |f| f.end_with?(release_notes_file) }
-      has_modified_app_store_strings = danger.git.modified_files.any? { |f| f.end_with?(store_strings_file) }
+    def check_release_notes_and_store_strings(release_notes_file:, po_file:)
+      has_modified_release_notes = danger.git.modified_files.any? { |f| f == release_notes_file }
+      has_modified_app_store_strings = danger.git.modified_files.any? { |f| f == po_file }
 
       return unless has_modified_release_notes && !has_modified_app_store_strings
 
-      report_message = "The `#{store_strings_file}` file should be updated if the editorialised release notes file `#{release_notes_file}` is being changed."
+      report_message = "The `#{po_file}` file should be updated if the editorialised release notes file `#{release_notes_file}` is being changed."
       message(report_message)
     end
 
--- a/lib/dangermattic/plugins/common/common_release_checks.rb
+++ b/lib/dangermattic/plugins/common/common_release_checks.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin performing a generic checks related to releases. Can be used directly or via specialised plugins as `AndroidReleaseCheck` and `IosReleaseCheck`.
+  class CommonReleaseChecks < Plugin
+    INTERNAL_RELEASE_NOTES = 'RELEASE-NOTES.txt'
+
+    # Check if certain files have been modified, returning a warning or failure message based on the branch type.
+    #
+    # @param file_comparison A closure used to compare modified file paths.
+    #   The closure should take a single argument, which is the path to a modified file,
+    #   and return true if the file matches the desired condition.
+    #   Example: `file_comparison = ->(file_path) { file_path.include?('app/') }`
+    #
+    # @param message [String] The message to display in the warning or failure output if the condition is met.
+    #
+    # @param on_release [Boolean] If true, the check will only run on release branches, otherwise on non-release branches.
+    #
+    # @param fail_on_error [Boolean] If true, a failure message will be displayed instead of a warning.
+    #
+    # @example Check if any modified file is under the 'app/' directory and emit a warning on release branches:
+    #   check_file_changed(file_comparison: ->(file_path) { file_path.include?('app/') },
+    #                      message: 'Some files in the "app/" directory have been modified. Please review the changes.',
+    #                      on_release: true
+    #   )
+    #
+    # @example Check if a specific file has been modified and emit a failure on non-release branches:
+    #   check_file_changed(file_comparison: ->(file_path) { file_path == 'path/to/file/DoNotChange.java' },
+    #                      message: 'The "DoNotChange.java" file has been modified. This change is not allowed on non-release branches.',
+    #                      on_release: false,
+    #                      fail_on_error: true)
+    #
+    def check_file_changed(file_comparison:, message:, on_release:, fail_on_error: false)
+      has_modified_file = all_modified_files.any?(&file_comparison)
+
+      should_be_changed = on_release ? release_branch? : !release_branch?
+      return unless should_be_changed && has_modified_file
+
+      if fail_on_error
+        failure(message)
+      else
+        warn(message)
+      end
+    end
+
+    # Check if the release notes and store strings files are correctly updated after a modification to the release notes.
+    #
+    # @param release_notes_file [String] The name of the release notes file that should be checked for modifications.
+    #   Example: 'metadata/release_notes.txt'
+    #
+    # @param store_strings_file [String] The name of the store strings file that should be checked for modifications.
+    #   Example: 'metadata/PlayStoreStrings.po'
+    #
+    # @param fail_on_error [Boolean] If true, a failure message will be emitted instead of a warning.
+    #
+    # @example Check if the release notes file 'release_notes.txt' is modified, and the 'PlayStoreStrings.po' file is not modified:
+    #   check_release_notes_and_store_strings(release_notes_file: 'release_notes.txt', store_strings_file: 'PlayStoreStrings.po', fail_on_error: false)
+    #
+    def check_release_notes_and_store_strings(release_notes_file:, store_strings_file:, fail_on_error: false)
+      has_modified_release_notes = danger.git.modified_files.any? { |f| f.end_with?(release_notes_file) }
+      has_modified_app_store_strings = danger.git.modified_files.any? { |f| f.end_with?(store_strings_file) }
+
+      return unless has_modified_release_notes && !has_modified_app_store_strings
+
+      message = "The #{File.basename(store_strings_file)} file must be updated any time changes are made to the release notes."
+
+      if fail_on_error
+        failure(message)
+      else
+        warn(message)
+      end
+    end
+
+    # Check if there are changes to the internal release notes file in the release branch and emit a warning.
+    def check_internal_release_notes_changed
+      warning = <<~WARNING
+        This PR contains changes to `RELEASE-NOTES.txt`.
+        Note that these changes won't affect the final version of the release notes as this version is in code freeze.
+        Please, get in touch with a release manager if you want to update the final release notes.
+      WARNING
+
+      check_file_changed(
+        file_comparison: ->(path) { path.end_with?(INTERNAL_RELEASE_NOTES) },
+        message: warning,
+        on_release: true,
+        fail_on_error: false
+      )
+    end
+
+    private
+
+    def all_modified_files
+      danger.git.added_files + danger.git.modified_files + danger.git.deleted_files
+    end
+
+    def release_branch?
+      danger.github.branch_for_base.start_with?('release/') || danger.github.branch_for_base.start_with?('hotfix/')
+    end
+  end
+end
--- a/lib/dangermattic/plugins/common/common_release_checks.rb
+++ b/lib/dangermattic/plugins/common/common_release_checks.rb
@@ -1,14 +1,35 @@
 # frozen_string_literal: true
 
 module Danger
-  # Plugin performing a generic checks related to releases. Can be used directly or via specialised plugins as `AndroidReleaseCheck` and `IosReleaseCheck`.
+  # Plugin to perform generic checks related to releases.
+  # It can be used directly or via the specialised plugins `AndroidReleaseCheck` and `IosReleaseCheck`.
+  #
+  # @example Checking if a specific file has changed on a release branch:
+  #          common_release_checks.check_file_changed(
+  #            file_comparison: ->(path) { path == 'metadata/full_release_notes.txt' },
+  #            message: 'Release notes have been modified on a release branch.',
+  #            on_release: true
+  #          )
+  #
+  # @example Checking if release notes and store strings have changed:
+  #          common_release_checks.check_release_notes_and_store_strings(
+  #            release_notes_file: 'metadata/release_notes.txt',
+  #            store_strings_file: 'metadata/PlayStoreStrings.po'
+  #          )
+  #
+  # @example Checking for changes in internal release notes:
+  #          common_release_checks.check_internal_release_notes_changed
+  #
+  # @see Automattic/dangermattic
+  # @tags util, process, release
+  #
   class CommonReleaseChecks < Plugin
     DEFAULT_INTERNAL_RELEASE_NOTES = 'RELEASE-NOTES.txt'
 
     # Check if certain files have been modified, returning a warning or failure message based on the branch type.
     #
-    # @param file_comparison A closure used to compare modified file paths.
-    #   The closure should take a single argument, which is the path to a modified file,
+    # @param file_comparison [Proc] Function used to compare modified file paths.
+    #   It should take a single argument, which is the path to a modified file,
     #   and return true if the file matches the desired condition.
     #   Example: `file_comparison = ->(file_path) { file_path.include?('app/') }`
     #
@@ -30,6 +51,7 @@ module Danger
     #                      on_release: false,
     #                      fail_on_error: true)
     #
+    # @return [void]
     def check_file_changed(file_comparison:, message:, on_release:, fail_on_error: false)
       has_modified_file = all_modified_files.any?(&file_comparison)
 
@@ -56,6 +78,7 @@ module Danger
     # @example Check if the release notes file 'release_notes.txt' is modified, and the 'PlayStoreStrings.po' file is not modified:
     #   check_release_notes_and_store_strings(release_notes_file: 'release_notes.txt', store_strings_file: 'PlayStoreStrings.po', fail_on_error: false)
     #
+    # @return [void]
     def check_release_notes_and_store_strings(release_notes_file:, store_strings_file:, fail_on_error: false)
       has_modified_release_notes = danger.git.modified_files.any? { |f| f.end_with?(release_notes_file) }
       has_modified_app_store_strings = danger.git.modified_files.any? { |f| f.end_with?(store_strings_file) }
@@ -82,6 +105,7 @@ module Danger
     # @example Checking for changes in a custom internal release notes file at a specific path:
     #   check_internal_release_notes_changed(release_notes_file: '/path/to/internal_release_notes.txt')
     #
+    # @return [void]
     def check_internal_release_notes_changed(release_notes_file: DEFAULT_INTERNAL_RELEASE_NOTES)
       warning = <<~WARNING
         This PR contains changes to `#{release_notes_file}`.
--- a/lib/dangermattic/plugins/common/common_release_checks.rb
+++ b/lib/dangermattic/plugins/common/common_release_checks.rb
@@ -54,7 +54,7 @@ module Danger
     def check_file_changed(file_comparison:, message:, on_release_branch:, fail_on_error: false)
       has_modified_file = git_utils.all_changed_files.any?(&file_comparison)
 
-      should_be_changed = on_release_branch ? release_branch? : !release_branch?
+      should_be_changed = (on_release_branch == release_branch?)
       return unless should_be_changed && has_modified_file
 
       if fail_on_error
--- a/lib/dangermattic/plugins/common/common_release_checks.rb
+++ b/lib/dangermattic/plugins/common/common_release_checks.rb
@@ -3,7 +3,7 @@
 module Danger
   # Plugin performing a generic checks related to releases. Can be used directly or via specialised plugins as `AndroidReleaseCheck` and `IosReleaseCheck`.
   class CommonReleaseChecks < Plugin
-    INTERNAL_RELEASE_NOTES = 'RELEASE-NOTES.txt'
+    DEFAULT_INTERNAL_RELEASE_NOTES = 'RELEASE-NOTES.txt'
 
     # Check if certain files have been modified, returning a warning or failure message based on the branch type.
     #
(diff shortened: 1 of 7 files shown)


lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      if fail_on_error
+        failure(message)
+      else
+        warn(message)
+      end
+    end
+
+    # Check if there are changes to the internal release notes file in the release branch and emit a warning.
+    def check_internal_release_notes_changed
```
AliSoftware: Would be nice to have this take the name of the `RELEASE-NOTES.txt` file, especially since we have on our (very-low-priority) roadmap to maybe rename those files to `CHANGELOG.md` to be more conventional (see pdnsEh-aj-p2) one day.

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,100 @@
...
+      warning = <<~WARNING
+        This PR contains changes to `RELEASE-NOTES.txt`.
```
AliSoftware: Better use `INTERNAL_RELEASE_NOTES` (or better, the value of that argument I suggested adding in the comment above) in this warning message, for consistency.

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,100 @@
...
+        Note that these changes won't affect the final version of the release notes as this version is in code freeze.
+        Please, get in touch with a release manager if you want to update the final release notes.
+      WARNING
+
+      check_file_changed(
+        file_comparison: ->(path) { path.end_with?(INTERNAL_RELEASE_NOTES) },
```
AliSoftware: You should not use `end_with?` on paths to check if their file name match. Because that will also match `'Components/Internal/PRIVATE_RELEASE_NOTES.txt'.end_with?('RELEASE_NOTES.txt')`.

Instead, to check if the file name matches, you should check `File.basename(path) == INTERNAL_RELEASE_NOTES`.

That being said, for this particular Danger rule, we want to only check one very specific file (the one maintained and manually edited by developers when they do a new PR), that is at a specific path (currently our `RELEASE-NOTES.txt` at the root of the repo). So for this rule we don't want to check _any_ arbitrary file named `RELEASE-NOTES.txt`, but only the one specified by the exact path of `RELEASE-NOTES.txt` **at the root of the repo**. So no need for `end_with?` or `File.basename == …` there.

lib/dangermattic/plugins/ios_release_check.rb
```diff
@@ -0,0 +1,35 @@
...
+module Danger
+  # Plugin for miscellaneous checks for iOS / macOS releases.
+  class IosReleaseCheck < Plugin
+    # Checks if an existing Core Data model has been edited in a release branch.
+    def check_core_data_model_changed
+      warning = 'Do not edit an existing Core Data model in a release branch unless it hasn\'t been released to testers yet. ' \
+                'Instead create a new model version and merge back to develop soon.'
+
+      common_release_checks.check_file_changed(
+        file_comparison: ->(path) { path.end_with?('.xcdatamodeld') },
```
AliSoftware: Use `File.extname(path) == '.xcdatamodeld'` instead

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,100 @@
...
+      return unless has_modified_release_notes && !has_modified_app_store_strings
+
+      message = "The #{File.basename(store_strings_file)} file must be updated any time changes are made to the release notes."
+
+      if fail_on_error
+        failure(message)
+      else
+        warn(message)
+      end
+    end
```
AliSoftware: I'm wondering if this old Peril Rule still has value nowadays given that:

1️⃣  it has always triggered a false positive in the 2-step process we follow during WP code freeze, where:
1. Usually the `metadata/release_notes.txt` gets changed during code freeze to be filled with the bare/raw/unedited release notes extracted from the top section of `RELEASE-NOTES.txt`, but at that point we don't want to update the `metadata/*StoreStrings.po` file with those because that would send unedited/raw release notes to GlotPress
2. And only when our freelance writer has massaged and wordsmithed a nice, editorialized release note text from those bare entries, do we update the `metadata/release_notes.txt` file with the nice copy, and only then run the lane that will update the `.po` file with that nice copy to be sent to GlotPress for translation

This means that historically, that Peril rule has always been triggered during step 1, and we've learned to ignore it as a false positive for the particular case of this step.

2️⃣  Also, with @oguzkocer working on migrating code freeze and other release operations to CI, I'm wondering if the future new process that will come from that will have an impact on those steps (1) and (2) above; in particular, if step (1) is done fully in CI via a `code_freeze/…` intermediate branch in the future, maybe there won't be a PR for step (1) anymore then? (in which case, if we will only have a PR in step (2), modifying both `release_notes.txt` and `*.po`, then maybe the rule will still have value now that the false positive from step 1 would be no more?

Interested to get the insights and thoughts from others @Automattic/apps-infrastructure , and in particular from folks who are doing Release Driving for any of the apps that rely on those `*.po` file + GlotPress (i.e. to my knowledge, at least WP/JP/WC… maybe others?)

AliSoftware (commented):
Preliminary review (I didn't get to the end of the PR as I'm EOD and have to go, but still wanted to submit what I reviewed so far)

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: Ah yes, I initially thought about it but considered that in such cases clients could just call `check_file_changed` directly. But looking at it again, it's a slightly lower level function and perhaps the default behaviour/message will be enough most of the times if you just want the check with another file. It won't hurt to add a parameter with a default value; I'll change it, thanks!

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: As a side note, @mokagio mentioned ( paaHJt-5bL-p2#comment-7740 ) they use the `Releases` label to avoid running these rules (thus avoiding the false positive warning) as a release manager.

iangmaia: Pinging again @Automattic/apps-infrastructure / @oguzkocer -- any thoughts to share on this?

oguzkocer: > 2️⃣ Also, with @oguzkocer working on migrating code freeze and other release operations to CI, I'm wondering if the future new process that will come from that will have an impact on those steps (1) and (2) above; in particular, if step (1) is done fully in CI via a code_freeze/… intermediate branch in the future, maybe there won't be a PR for step (1) anymore then? (in which case, if we will only have a PR in step (2), modifying both release_notes.txt and *.po, then maybe the rule will still have value now that the false positive from step 1 would be no more?

Regardless of whether the code freeze happens in CI or in our local environment, there has to be a PR to merge the created release branch to `trunk`. So, the release management in CI changes shouldn't impact this.

---

In my experience, whenever I saw this warning it was always a false positive. That might be because release managers know when the `.po` files should be updated, but when we start rotating product developers to do releases, they would not have this knowledge. Our release scenarios are very detailed, so this is unlikely to be a problem, but it may still be useful to share this info in PRs just in case.

So, if we can work on the messaging and make this an info rather than an error or a warning, I think that'd probably provide the most value without getting in the way.

> message = "The #{File.basename(store_strings_file)} file must be updated any time changes are made to the release notes."

Maybe instead of this, it should say `.po` file `should` be updated `if the editorialized release notes are changed`. Or we can be more verbose about it when to update the `.po` file and when not to do it. 🤷

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,134 @@
...
+    #
+    # @param fail_on_error [Boolean] If true, a failure message will be emitted instead of a warning.
+    #
+    # @example Check if the release notes file 'release_notes.txt' is modified, and the 'PlayStoreStrings.po' file is not modified:
+    #   check_release_notes_and_store_strings(release_notes_file: 'release_notes.txt', store_strings_file: 'PlayStoreStrings.po', fail_on_error: false)
+    #
+    # @return [void]
+    def check_release_notes_and_store_strings(release_notes_file:, store_strings_file:, fail_on_error: false)
+      has_modified_release_notes = danger.git.modified_files.any? { |f| f.end_with?(release_notes_file) }
+      has_modified_app_store_strings = danger.git.modified_files.any? { |f| f.end_with?(store_strings_file) }
```
AliSoftware: Better check the `File.basename` rather than `f.end_with?` because otherwise `f = 'previous_release_notes.txt'` would match `end_with?('release_notes.txt')` for example.

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,134 @@
...
+
+      return unless has_modified_release_notes && !has_modified_app_store_strings
+
+      message = "The #{File.basename(store_strings_file)} file must be updated any time changes are made to the release notes."
```
AliSoftware: I like Oguz's new wording suggestion from https://github.com/Automattic/dangermattic/pull/9#discussion_r1298820641, maybe we should use it instead.

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,134 @@
...
+        on_release: true,
+        fail_on_error: false
+      )
+    end
+
+    private
+
+    def all_modified_files
+      danger.git.added_files + danger.git.modified_files + danger.git.deleted_files
+    end
```
AliSoftware: This seems common enough to be added too a helper method that could be shared across multiple plugins of this repo?

Also, back when I wrote a couple of Danger rules, I remember that it reports renamed files in a tricky way, i.e. something like when a file is renamed, `danger.git.renamed_files` is set with the list of `{ old_name => new_name }`… but the name used in `danger.git.modified_files` in that case might not be the one you expect (i.e. the old name and not the new name).
Which could lead to red herrings and false positives in our logic (e.g. when we test if file `b.txt` was changed, but what happened in reality is that we renamed `a.txt` to `b.txt`… but `danger.git.modified_files` will list `a.txt`, not `b.txt` so our rule will not detect that `b.txt` was modified…

This is something to address is a separate PR, e.g. when you'll extract that method in some common helper; at that point you might experiment with some real data (e.g. create a real PR that renames some changes and have a `Dangerfile` print the content of `danger.git.added_files.inspect`, `danger.git.modified_files.inspect`, `danger.git.deleted_files.inspect` and `danger.git.renamed_files.inspect`, to see what's returned in reality, so that you can write specs and fixtures for your helper method that reflect that reality, and then make sure it takes those cases into account…?

lib/dangermattic/plugins/ios_release_check.rb
```diff
@@ -0,0 +1,54 @@
...
+        on_release: true
+      )
+    end
+
+    # Checks if the Localizable.strings file has been modified on a regular branch, emiting a warning if that's the case.
+    #
+    # @return [void]
+    def check_modified_localizable_strings
+      common_release_checks.check_file_changed(
+        file_comparison: ->(path) { path.end_with?('Resources/en.lproj/Localizable.strings') },
```
AliSoftware: `end_with?` is not an appropriate test to check paths, because that could match a partial name of a directory in the start of the string you're testing.

Besides, I'm not sure all our repos have their `en.lproj` inside `Resources`? (I think most do… but is that an absolute rule for _all_ our repos?

Finally, since this rule seems to be aimed at repositories that make use of `genstrings` (likely via [our `ios_generate_strings_file_from_code` action from `release-toolkit`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_generate_strings_file_from_code.rb)), I think we should also warn if _any_ `Localizable.strings` file is modified, because:
 - The `en.lproj` file should only be updated when we run `genstrings` (on release branch, right after code freeze), but never updated manually
 - The other `*.lproj` files should only be updated when we download translations from GlotPress, which is also done on release branch, typically just before submission

Tumblr is a bit different, as it doesn't (yet) use `genstrings` to generate the `en.lproj` file (instead, developers update the `en.lproj/Localizable.strings` file manually when they need to add a new string). Yet, developers should not update any other `*.lproj/Localizable.strings` file manually themselves, as these translation files should only be updated on release branch when we download translations from OneSky.

All this means that:
 - For repos that use `genstrings`, neither `en.lproj` nor any other `*.lproj` file should be edited manually (but only updated by our automation on the release branch)
 - For Tumblr repo (or any other that don't use `genstrings`), editing `en.lproj/Localizable.strings` manually is allowed if on `develop` branch (or more precisely, if on any PRs which is NOT targeting `release/`), but editing any other `*.lproj/Localizable.strings` manually is not allowed, and should only be done via our automation on the `release/` branch

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: Though at the same time, in this case given I'd expect users to specify a path, one could specify `metadata/release_notes.txt` and it would match `old/release_notes.txt`. So perhaps a `==` match would be better 🤔

AliSoftware: Indeed. So better just `==` the two strings and expect the caller to provide the _exact_ path relative to the root of the client project's repo.

AliSoftware: Btw, `store_strings_file` isn't super explicit and fitting of a variable name. I'd instead suggest `po_file` or something similar?

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: **On separate branches**, I've already turned the class `GitUtils` into a [common plugin and added this util method there](https://github.com/Automattic/dangermattic/blob/17ab16e6c2867b229e080debdd283117a7651229/lib/dangermattic/plugins/common/git_utils.rb).
Later I've also [renamed](https://github.com/Automattic/dangermattic/blob/24943497dfd32fb1ee7d573b4555c67da268bc7a/lib/dangermattic/plugins/common/git_utils.rb) it to `all_changed_files` because the name per se was confusing (similar to the `modified_files` used by the Git API as well).

The case you mentioned about renames, tbh I'm not sure yet, but that's a good point for me to test as well 👍

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: Thanks for the suggestion, @oguzkocer (sorry for the delay)! I've slightly changed the message and the type of report (just a regular `message` now): 29d8949627a2f60827c6497b85bd0662c228ed2e

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: Thanks for the suggestions -- I have changed a few things on this method here: e687615ca59e7c026dd12188cec5df1413139559

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: Yep, changed here: https://github.com/Automattic/dangermattic/commit/29d8949627a2f60827c6497b85bd0662c228ed2e

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -70,28 +69,21 @@ def check_file_changed(file_comparison:, message:, on_release:, fail_on_error: f
...
 
-      message = "The #{File.basename(store_strings_file)} file must be updated any time changes are made to the release notes."
-
-      if fail_on_error
-        failure(message)
-      else
-        warn(message)
-      end
+      report_message = "The `#{po_file}` file should be updated if the editorialised release notes file `#{release_notes_file}` is being changed."
+      message(report_message)
```
AliSoftware: 👍

lib/dangermattic/plugins/ios_release_check.rb
iangmaia: Good points, thanks -- this will help to make the check a bit more general (it was first just based on what we had on Peril).
I've made a few changes here e687615ca59e7c026dd12188cec5df1413139559.
If I got it right, I've also added two separate checks to cover the "Tumblr case", 100847145854005297e37e78988d8848daad46ea and 67caccf9b115134ac317782edf94e5e80aab3a54.

iangmaia:
@AliSoftware, do you have further comments in this PR?

lib/dangermattic/plugins/ios_release_check.rb
```diff
@@ -0,0 +1,79 @@
...
+      )
+    end
+
+    # Checks if any Localizable.strings file has been modified on a release branch, otherwise reporting a warning.
+    #
+    # @return [void]
+    def check_modified_localizable_strings_on_release
+      common_release_checks.check_file_changed(
+        file_comparison: ->(path) { File.basename(path) == LOCALIZABLE_STRINGS_FILE },
+        message: "The `#{LOCALIZABLE_STRINGS_FILE}` files should only be updated on release branches, when the translations are downloaded.",
```
AliSoftware:
```suggestion
        message: "The `#{LOCALIZABLE_STRINGS_FILE}` files should only be updated on release branches, when the translations are downloaded by our automation.",
```
Just to make it clear that even on release branches, they should not be changed manually.

lib/dangermattic/plugins/ios_release_check.rb
```diff
@@ -0,0 +1,79 @@
...
+        on_release: true
+      )
+    end
+
+    # Checks if a translation file (*.lproj/Localizable.strings) has been modified on a release branch, otherwise reporting a warning.
+    #
+    # @return [void]
+    def check_modified_translations_on_release_branch
+      common_release_checks.check_file_changed(
+        file_comparison: ->(path) { !path.end_with?(BASE_STRINGS_FILE) && File.basename(path) == LOCALIZABLE_STRINGS_FILE },
```
AliSoftware: Same feedback as before wrt the use of `end_with?` to compare file paths.

Prefer using `File.basename(path) != BASE_STRINGS_FILE` instead of `!path.end_with?(BASE_STRINGS_FILE)` to avoid matching unexpected files like `fi-en.lproj/Localizable.strings` or whatnot (not a valid locale code here, but you see my point)

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,122 @@
...
+    #                      on_release: true)
+    #
+    # @example Check if a specific file has been modified and emit a failure on non-release branches:
+    #   check_file_changed(file_comparison: ->(file_path) { file_path == 'path/to/file/DoNotChange.java' },
+    #                      message: 'The "DoNotChange.java" file has been modified. This change is not allowed on non-release branches.',
+    #                      on_release: false,
+    #                      fail_on_error: true)
+    #
+    # @return [void]
+    def check_file_changed(file_comparison:, message:, on_release:, fail_on_error: false)
```
AliSoftware: I wonder if it wouldn't be clearer to use `on_release_branch` rather than just `on_release`? wdyt?

lib/dangermattic/plugins/common/common_release_checks.rb
```diff
@@ -0,0 +1,122 @@
...
+      has_modified_file = git_utils.all_changed_files.any?(&file_comparison)
+
+      should_be_changed = on_release ? release_branch? : !release_branch?
```
AliSoftware: Might be more readable if you wrote this boolean expression like this? 🤷 
```suggestion
      should_be_changed = (on_release == release_branch?)
```

lib/dangermattic/plugins/ios_release_check.rb
```diff
@@ -0,0 +1,79 @@
...
+        file_comparison: ->(path) { File.extname(path) == '.xcdatamodeld' },
+        message: warning,
+        on_release: true
+      )
+    end
+
+    # Checks if any Localizable.strings file has been modified on a release branch, otherwise reporting a warning.
+    #
+    # @return [void]
+    def check_modified_localizable_strings_on_release
```
AliSoftware, line 39: I think it would be valuable to implement the same rule for Android (on `res/values-*/strings.xml` files) for the same reason as iOS?

lib/dangermattic/plugins/ios_release_check.rb
iangmaia: Though `File.basename(path) != BASE_STRINGS_FILE` will always be true, even when `path = en.lproj/Localizable.strings`, since `File.basename(path)` will return `Localizable.strings`, right?
Perhaps just using `path != BASE_STRINGS_FILE` is best, but I'm not sure if `path` can have other elements that would break the check (like `./` at the start).

Haven't tested it, but what if we do
```ruby
BASE_STRINGS_FILE_PATTERN = "*/en.lproj/#{LOCALIZABLE_STRINGS_FILE}"
[...]
{ !File.fnmatch(BASE_STRINGS_FILE_PATTERN, path) && File.basename(path) == LOCALIZABLE_STRINGS_FILE }
```

AliSoftware: Maybe a helper method that would check if `path == LOCALIZABLE_STRINGS_FILE || path.end_with?('/' + LOCALIZABLE_STRINGS_FILE.delete_prefix('/'))`, to account for both the case of if `LOCALIZABLE_STRINGS_FILE` happen to be rooted at the root of the repo (hence not being in any subdirectory, and thus not having a leading `/` if it's already the top dir) and adding a `/` prefix if not present already for other cases…

Though I think implementing such logic like that could still make us miss some edge cases (like paths containing `..` and stuff that we should maybe run `File.realpath` or `Pathname.cleanpath` on them first. Actually, I'm wondering if there isn't any method from `Pathname` (which is a bit feature-rich compared to `File.…` methods when manipulating paths) that would already be able to do that for us?

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: Much better 👍 Changed here: da81323cd52c697cef850464fbc3d0bbf8650c43

lib/dangermattic/plugins/ios_release_check.rb
iangmaia: Digging the docs, maybe something like this would work:

```
base_path_name_components = Pathname.new(BASE_STRINGS_FILE).each_filename.to_a
Pathname.new(path).each_filename.to_a.last(base_path_name_components.length) == base_path_name_components
```

AliSoftware: Wfm. To be extracted in a helper method to give it a nice name.

lib/dangermattic/plugins/common/common_release_checks.rb
iangmaia: Yeah, I think it reads better. Changed here: 4b83ceabe6e73dec3a0d6de3c39e553e24704a1a

lib/dangermattic/plugins/ios_release_check.rb
iangmaia: Agreed -- this wasn't originally in Peril iinm, but I do remember you mentioning this check in another discussion and I even started implementing it, I might have it stashed. ~Do you think it is a requirement for this PR to land or I can add a task in the iteration project to add it? It's not a lot in any case.~ Nvm, I've added it here: e32031c6483b77e8e2c2c457000e6dcbde433018

lib/dangermattic/plugins/ios_release_check.rb
iangmaia: Updated it: 0b8318e6ecc51e32618c600db95707367d24c2ec

lib/dangermattic/plugins/ios_release_check.rb
```diff
@@ -0,0 +1,88 @@
...
+      )
+    end
+
+    private
+
+    def base_strings?(path:)
+      base_strings_path_components = Pathname.new(BASE_STRINGS_FILE).each_filename.to_a
+      path_components = Pathname.new(path).each_filename.to_a
+
+      base_strings_path_components == path_components
```
AliSoftware: Missing the `.last(base_path_name_components.length)` that you mentioned in https://github.com/Automattic/dangermattic/pull/9#discussion_r1394751066, otherwise this basically would end up being the same as comparing `path == BASE_STRINGS_FILE` directly (i.e. splitting each into components wouldn't do much… not even resolving potential `../` as you don't call `.cleanpath`)

Method could use some YARD doc as well.

iangmaia: Oh, you're right -- I thought for a moment today this would be equivalent but indeed there are other scenarios to cover. Amended in e1e80a10b700ec3e05b6d2256594f4c4e347d6fe.

lib/dangermattic/plugins/ios_release_check.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    # Checks if a given path corresponds to the base (English) strings file, en.lproj/Localizable.strings.
+    #
+    # @return [Boolean] true if path is the base strings file
+    def base_strings_file?(path:)
+      base_strings_path_components = Pathname.new(BASE_STRINGS_FILE).each_filename.to_a
+      path_components = Pathname.new(path).each_filename.to_a
+
+      base_strings_path_components == path_components.last(base_strings_path_components.length)
+    end
```
AliSoftware, line 89: I think in the future this method could be generalized to some `def matches_relative_path?` or something, in the collection of utility method that you're working on, so that every time we need a path comparison where you used to rely on `end_with?` in any of your rules you'd now be able to use that more resilient helper instead.

That could be for a future PR though.

AliSoftware (approved):

lib/dangermattic/plugins/ios_release_check.rb
iangmaia: Yep, I had the same thought. Will look into when I move some more common code to utils classes 👍

PR #10, opened 2023-07-21
Plugins for Podfile and Android strings checks

iangmaia:
This PR introduces the plugins `AndroidStringsChecks`, `PodfileChecks` and `GitUtils` (now as a plugin as opposed to a simple class).
These plugins brings over two checks from our Peril rules: 
- [This check](https://github.com/Automattic/peril-settings/blob/7f7e674c099cff6c0af803f0941f3353b2b395e2/org/pr/ios-macos.ts#L21C6-L21C6) on [ios-macos.ts](https://github.com/Automattic/peril-settings/blob/master/org/pr/ios-macos.ts) that verifies that we're not referencing commits directly in a Podfile.
  - There are two versions of a similar check: one identical to `ios-macos.ts`, which loads the repo's `Podfile.lock` and one performing the same check on a PR diff. I implemented the latter one so it is consistent with the rest of the checks we have in Dangermattic so far, but perhaps this could be a choice from each repo when applicable (PR check vs. full project check). 
- [This check](https://github.com/Automattic/peril-settings/blob/7f7e674c099cff6c0af803f0941f3353b2b395e2/org/pr/android.ts#L19C11-L19C11) on [android.ts](https://github.com/Automattic/peril-settings/blob/master/org/pr/android.ts) that warns if a string in a strings.xml file is referencing another string resource.

They were both in the general iOS/Android rules file and didn't really fit into the [Releases Checks plugin](https://github.com/Automattic/dangermattic/pull/9).

Both plugins use the new `GitUtils` plugin, which was [originally](https://github.com/Automattic/dangermattic/blob/8957a5eb048af2dc05d63848988e8d8476bf3209/lib/dangermattic/plugins/utils/git_utils.rb) a plain util class, but I decided to experiment with it being a Danger plugin that can be used either directly or by other plugins. The main advantages of it being a plugin are:
- Being directly accessible to any other plugin or `Dangerfile`, automatically being instantiated by Danger and visible as `git_utils`
- Having access to `danger`, `danger.git` and other common Danger components, so these don't need to be passed as parameter or injected, simplifying the setup.

_[**update**: instead of naming "miscellaneous" plugin checks, using specific plugin names instead ([ref](https://github.com/Automattic/dangermattic/pull/10#discussion_r1282039042))]_


lib/dangermattic/plugins/android_misc_checks.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin for miscellaneous checks for Android.
+  class AndroidMiscChecks < Plugin
```
AliSoftware: I'm not a fan of this "fits-it-all bucket" kind of thing. This can easily become a place to dump everything blindly over time and make it a bit of a mess with lots of unrelated checks.

> They were both in the general iOS/Android rules file and didn't really fit into the https://github.com/Automattic/dangermattic/pull/9, but so far they're not really close to anything else, therefore miscellaneous checks. As new plugins added with time, possibly we can group them in another way if needed.

I think it'd make more sense to go the other way around, i.e. make the plugin have a name that relates to a specific area, like `AndroidStringsChecks`, that can evolve later to add more checks in that same area (I can already imagine checks for ensuring that a string with an existing key didn't have its value mutated (p1690456266010749/1690454415.738839-slack-CRF8CPYAK).
And if over time we realize that our grouping of the checks could have been done in a different way and split into different plugins, we could always refactor (and at that time we'll then have enough examples of plugins and checks in our toolset to reason about a better grouping if needed)

lib/dangermattic/plugins/android_misc_checks.rb
```diff
@@ -0,0 +1,18 @@
...
+    # Check if translatable strings reference another string resource in 'strings.xml' files in a pull request.
+    def check_strings_do_not_refer_resource
+      warning_message = "This PR adds a translatable entry which references another string resource; this usually causes issues with translations.\n" \
+                        'Please make sure to set the `translatable="false"` attribute:'
+
+      git_utils.check_added_diff_lines(
+        file_selector: ->(path) { File.basename(path) == 'strings.xml' },
+        line_matcher: ->(line) { line.include?('@string/') && !line.include?('translatable="false"') },
```
AliSoftware: I wonder if it wouldn't be better for this to be a lint check, i.e. done by a lint tool that runs on the whole file on every CI run and parses the whole file using an XML parser, rather than a Danger check only looking for the changed files?

lib/dangermattic/plugins/ios_misc_checks.rb
```diff
@@ -0,0 +1,55 @@
...
+
+require 'yaml'
+
+module Danger
+  PODFILE_LOCK = 'Podfile.lock'
+  PODFILE_LOCK_DEPENDENCIES_ENTRY = 'DEPENDENCIES'
+  DEFAULT_PODFILE_LOCK_PATH = './Podfile.lock'
+
+  # Plugin for miscellaneous iOS checks.
+  class IosMiscChecks < Plugin
```
AliSoftware: Same feedback as above, I noted your paragraph in the PR description but I still think we should use a better name, like `PodfileChecks`?

lib/dangermattic/plugins/android_misc_checks.rb
iangmaia: > I think it'd make more sense to go the other way around, i.e. make the plugin have a name that relates to a specific area, like `AndroidStringsChecks`, that can evolve later to add more checks in that same area (I can already imagine checks for ensuring that a string with an existing key didn't have its value mutated)

I agree, it is too easy to have a "basement" plugin where you just throw stuff you don't know where else to put. I'll move things around to specific plugins for now. Good suggestion!

lib/dangermattic/plugins/android_misc_checks.rb
iangmaia: I had the same question myself; that came to mind after porting the Peril rule in this PR for the `Podfile`, *which checks the entire file instead of only the diff*. For the strings check, it would also feel more natural to have a full lint check.

For now I just brought what we had in the Peril rule. Tbf, I do like the idea of a collection of custom lint rules, but that seems like a separate discussion (though it would be nice to standardise the way we run lint in projects the same way we standardise the Danger checks). Perhaps some Danger checks could, after all, do full project runs (like the aforementioned Podfile one does).

lib/dangermattic/plugins/podfile_checks.rb
```diff
@@ -0,0 +1,68 @@
...
+    # @example Checking the default Podfile.lock:
+    #   check_podfile_does_not_have_commit_references
+    #
+    # @example Checking a custom Podfile.lock at a specific path:
+    #   check_podfile_does_not_have_commit_references(podfile_lock_path: '/path/to/Podfile.lock')
+    #
+    # @return [void]
+    def check_podfile_does_not_have_commit_references(podfile_lock_path: DEFAULT_PODFILE_LOCK_PATH)
+      podfile_lock_contents = File.read(podfile_lock_path)
+      podfile_lock_data = YAML.load(podfile_lock_contents)
```
mokagio, line 36: 👏 for parsing the YAML! One earlier implementation somewhere was using RegExs but this approach is much more robust.

spec/android_strings_checks_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+          WARNING
+
+          expected_warning3 = <<~WARNING
+            This PR adds a translatable entry which references another string resource; this usually causes issues with translations.
+            Please make sure to set the `translatable="false"` attribute:
+            File `#{ptbr_strings_xml}`:
+            ```diff
+            +  <string name="popup_title">@string/app_name_title</string>
+            ```
+          WARNING
```
mokagio, line 129: Have you considered passing a block to generated the warning at the plugin level?

It would make testing this more straightforward because we could simply verify it interpolates the diff correctly. 

Something like in the PR size check, but where the evaluation happens at runtime, not callsite.

https://github.com/Automattic/dangermattic/blob/trunk/lib/dangermattic/plugins/pr_size_checker.rb#L49

Just an idea, though. I can see how it might be a bit forced, code written just for the sake of making the testing smoother.

lib/dangermattic/plugins/common/git_utils.rb
```diff
@@ -0,0 +1,142 @@
...
+  #
+  # @example Check if there is a "TODO" in Ruby files:
+  #          git_utils.check_added_diff_lines(
+  #            file_selector: ->(path) { path.end_with?('.rb') },
+  #            line_matcher: ->(line) { line.include?('TODO') },
+  #            message: 'Found a TODO in a Ruby file'
+  #          )
+  #
+  # @example Get added lines from a diff patch:
+  #          added_lines = added_lines(diff_patch: diff_patch)
```
mokagio: Should this and the other examples call `git_utils`, too?

```suggestion
  #          added_lines = git_utils.added_lines(diff_patch: diff_patch)
```

lib/dangermattic/plugins/podfile_checks.rb
```diff
@@ -0,0 +1,68 @@
...
+      )
+    end
+
+    private
+
+    def podfile_lock_commit_reference?(podfile_line:)
+      podfile_line.match?(/\(from `\S+`, commit `\S+`\)/)
+    end
+  end
+end
```
mokagio, line 68: Heads up: I can see this rule needing to change from "do not use commit references" to "do not use branch references". Or, maybe not change, but be expanded to support both.

The rationale is that in the context of not requiring developers to create beta builds for the changes they make to the libraries, it might be safer to ask to reference commits rather than branches, as commits are locked, but branches keep updating (and might be deleted in the remote).

For comparison, some Android dependencies like [Flux](https://github.com/wordpress-mobile/WordPress-FluxC-Android) are published automatically using the commit as the identifier.

lib/dangermattic/plugins/android_strings_checks.rb
```diff
@@ -0,0 +1,28 @@
...
+  #
+  class AndroidStringsChecks < Plugin
+    # Check if translatable strings reference another string resource in 'strings.xml' files in a pull request.
+    #
+    # @return [void]
+    def check_strings_do_not_refer_resource
+      warning_message = "This PR adds a translatable entry which references another string resource; this usually causes issues with translations.\n" \
+                        'Please make sure to set the `translatable="false"` attribute:'
+
+      git_utils.check_added_diff_lines(
```
mokagio, line 21: `git_utils` is available thanks to the change done in `danger_plugin.rb` above.

I wonder what would happen if someone used this "sub-plugin" directly?

lib/dangermattic/plugins/common/git_utils.rb
iangmaia: Yes! Added here: 0121c04c3d2e9eb57fc8f082806a07a2eb30e728

lib/dangermattic/plugins/android_strings_checks.rb
iangmaia: I think that shouldn't be a problem, and is actually an upside. I already see a few other utilities that can be moved to `git_utils` so that they don't need to be repeated across several `Dangerfile`s.

lib/dangermattic/plugins/podfile_checks.rb
iangmaia: > it might be safer to ask to reference commits rather than branches, as commits are locked, but branches keep updating (and might be deleted in the remote).

Good point, I think extending the plugin to support also check for not using branch references is a nice next step. I've added it to the [tasks backlog](https://github.com/orgs/Automattic/projects/775/views/1?pane=issue&itemId=42548719).

spec/android_strings_checks_spec.rb
iangmaia: Hmm yeah I'm trying to think other advantages (other than making testing smoother) but not sure, I wonder if it wouldn't make the function a bit more exotic than it should be 🤔  [sorry for the delay on the reply, I was going back pending PRs and realised I never got back to this one].

mokagio: No worries 👍 

I can't make a strong case for this, happy to leave it.

lib/dangermattic/plugins/android_strings_checks.rb
mokagio: > I already see a few other utilities that can be moved to `git_utils` so that they don't need to be repeated across several `Dangerfile`s.

Sure 👍 

Would the import work though, if one fetches the sub-plugin only instead of the whole Dangermattic? I honestly have no idea 😅

mokagio (approved):

lib/dangermattic/plugins/android_strings_checks.rb
iangmaia: Once you add the `Dangermattic` Gem, all plugins are instantiated and made available -- it's all magic from Danger 🙂.
It is indeed possible to fetch only a certain plugin in isolation, but then I **think** it would fail 🤔

PR #12, opened 2023-07-31
Plugin for checking the PR size

iangmaia:
This PR adds a plugin to perform checks on the PR diff size and body length.

It is based on the following Peril rules:
- https://github.com/Automattic/peril-settings/blob/master/org/pr/ios-diff-size.ts
- https://github.com/Automattic/peril-settings/blob/master/org/pr/android-diff-size.ts

The checks were made to cover a more general use case, such as ignoring certain files (for example, so that test files wouldn't count against a maximum size).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,9 @@ require:
 AllCops:
   NewCops: enable
 
+Style/HashSyntax:
+  EnforcedShorthandSyntax: either
+
 ########## Gemspec Rules
 
 # Disabling this as all the gem publishing will be done within CI and doesn't allow for user input such as an MFA code.
@@ -57,3 +60,6 @@ RSpec/ExampleLength:
 # Same for number of let/subject
 RSpec/MultipleMemoizedHelpers:
   Enabled: false
+
+RSpec/MultipleExpectations:
+  Max: 2
--- a/lib/dangermattic/plugins/pr_size_checker.rb
+++ b/lib/dangermattic/plugins/pr_size_checker.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin to check the size of a Pull Request content and text body.
+  class PRSizeChecker < Plugin
+    DEFAULT_MAX_DIFF_SIZE = 500
+    DEFAULT_DIFF_SIZE_MESSAGE = 'Please keep the Pull Request small, breaking it down into multiple ones if necessary'
+    DEFAULT_MIN_PR_BODY_MESSAGE = 'Please provide a summary of the changes in the Pull Request description'
+    DEFAULT_MIN_PR_BODY = 10
+
+    # Check the size of the PR diff against a specified maximum size.
+    #
+    # @param file_selector [Proc] Optional closure to filter the files in the diff to be used for size calculation.
+    # @param type [:insertions, :deletions, :all] The type of diff size to check. (default: :all)
+    # @param max_size [Integer] The maximum allowed size for the diff. (default: DEFAULT_MAX_DIFF_SIZE)
+    # @param message [String] The message to display if the diff size exceeds the maximum. (default: DEFAULT_DIFF_SIZE_MESSAGE)
+    # @param fail_on_error [Boolean] If true, fail the PR check when the diff size exceeds the maximum (default: false).
+    def check_diff_size(file_selector: nil, type: :all, max_size: DEFAULT_MAX_DIFF_SIZE, message: DEFAULT_DIFF_SIZE_MESSAGE, fail_on_error: false)
+      case type
+      when :insertions
+        if insertions_size(file_selector: file_selector) > max_size
+          fail_on_error ? failure(message) : warn(message)
+        end
+      when :deletions
+        if deletions_size(file_selector: file_selector) > max_size
+          fail_on_error ? failure(message) : warn(message)
+        end
+      when :all
+        if diff_size(file_selector: file_selector) > max_size
+          fail_on_error ? failure(message) : warn(message)
+        end
+      end
+    end
+
+    # Check the size of the Pull Request description (PR body) against a specified minimum size.
+    #
+    # @param min_length [Integer] The minimum allowed length for the PR body. (default: DEFAULT_MIN_PR_BODY)
+    # @param fail_on_error [Boolean] If true, fail the PR check when the PR body length is too small. (default: false)
+    def check_pr_body(min_length: DEFAULT_MIN_PR_BODY, message: DEFAULT_MIN_PR_BODY_MESSAGE, fail_on_error: false)
+      return if danger.github.pr_body.length > min_length
+
+      fail_on_error ? failure(message) : warn(message)
+    end
+
+    # Calculate the total size of insertions in modified files that match the file selector.
+    #
+    # @param file_selector [Proc] Select the files to be used for the insertions calculation.
+    # @return [Integer] The total size of insertions in the selected modified files.
+    def insertions_size(file_selector:)
+      return danger.git.insertions unless file_selector
+
+      filtered_files = all_modified_files.select(&file_selector)
+      filtered_files.sum { |file| danger.git.info_for_file(file)[:insertions].to_i }
+    end
+
+    # Calculate the total size of deletions in modified files that match the file selector.
+    #
+    # @param file_selector [Proc] Select the files to be used for the deletions calculation.
+    # @return [Integer] The total size of deletions in the selected modified files.
+    def deletions_size(file_selector:)
+      return danger.git.deletions unless file_selector
+
+      filtered_files = all_modified_files.select(&file_selector)
+      filtered_files.sum { |file| danger.git.info_for_file(file)[:deletions].to_i }
+    end
+
+    # Calculate the total size of changes (insertions and deletions) in modified files that match the file selector.
+    #
+    # @param file_selector [Proc] Select the files to be used for the total insertions and deletions calculation.
+    # @return [Integer] The total size of changes in the selected modified files.
+    def diff_size(file_selector:)
+      return danger.git.lines_of_code unless file_selector
+
+      filtered_files = all_modified_files.select(&file_selector)
+      filtered_files.sum { |file| danger.git.info_for_file(file)[:deletions].to_i + danger.git.info_for_file(file)[:insertions].to_i }
+    end
+
+    private
+
+    def all_modified_files
+      danger.git.added_files + danger.git.modified_files + danger.git.deleted_files
+    end
+  end
+end
--- a/lib/dangermattic/plugins/pr_size_checker.rb
+++ b/lib/dangermattic/plugins/pr_size_checker.rb
@@ -4,9 +4,9 @@ module Danger
   # Plugin to check the size of a Pull Request content and text body.
   class PRSizeChecker < Plugin
     DEFAULT_MAX_DIFF_SIZE = 500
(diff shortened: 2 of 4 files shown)


spec/pr_size_checker_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+
+            type_hash[type]
+          end
+
+          it 'returns a warning when using default parameters in a PR that has larger diff than the default maximum' do
+            allow(@plugin.git).to receive(diff_counter_for_type).and_return(501)
+
+            @plugin.check_diff_size(type: type)
+
+            expect(@dangerfile.status_report[:warnings]).to eq ['Please keep the Pull Request small, breaking it down into multiple ones if necessary']
```
mokagio: 🤔 I like the suggestion to split the PR. This warning doesn't define what "small" means, or rather why this is not small.

Have you considered interpolating the lines threshold in the message, e.g. [like the iOS Peril check does](https://github.com/Automattic/peril-settings/blob/7f7e674c099cff6c0af803f0941f3353b2b395e2/org/pr/ios-diff-size.ts#L16)?

mokagio: I guess one should check the errors, too, for consistency with the other examples and thoroughness?

```suggestion
            expect(@dangerfile.status_report[:warnings]).to eq ['Please keep the Pull Request small, breaking it down into multiple ones if necessary']
            expect(@dangerfile.status_report[:errors]).to be_empty
```

spec/pr_size_checker_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+          it 'does nothing when using default parameters in a PR that has smaller diff than the default maximum' do
+            allow(@plugin.git).to receive(diff_counter_for_type).and_return(499)
+
+            @plugin.check_diff_size(type: type)
+
+            expect(@dangerfile.status_report[:errors]).to be_empty
+            expect(@dangerfile.status_report[:warnings]).to be_empty
+          end
+
+          it 'returns an error using a custom message and a custom pr body size' do
```
mokagio: For consistency:

```suggestion
          it 'returns an error using a custom message and a custom PR body size' do
```

spec/pr_size_checker_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+            allow(@plugin.git).to receive(diff_counter_for_type).and_return(600)
+
+            custom_message = 'diff size custom message'
+            @plugin.check_diff_size(type: type, max_size: 599, message: custom_message, fail_on_error: true)
+
+            expect(@dangerfile.status_report[:errors]).to eq [custom_message]
+          end
```
mokagio: This test looks like it's asserting two distinct behaviors: that a custom message is used when given, and that when we `fail_on_error: true` an error is reported.

Would you consider splitting into multiple `it`s? I'm thinking

- it reports a warning with a custom message when one is given
- it reports a warning when fail on error is false
- it reports an error when fail on error is true
- it reports an error with a custom message when one is given and fail on error is true (optional)

spec/pr_size_checker_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+            @plugin.check_diff_size(
+              file_selector: ->(path) { File.dirname(path).start_with?('src/test/java') },
+              type: type,
+              max_size: max_sizes[0]
+            )
+
+            expect(@dangerfile.status_report[:warnings]).to eq ['Please keep the Pull Request small, breaking it down into multiple ones if necessary']
+          end
+
+          it 'returns a warning when using a files filter that will result in a diff that is too large, with a custom error' do
```
mokagio: Have you considered using a `context` here?

```ruby
context 'when using a files filter that will regard the diff as too large' do
  it 'reports a warning' do ...
  it 'reports a warning with a given custom message' do ...
 ...
```

spec/pr_size_checker_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+            type_hash = {
+              insertions: :insertions,
+              deletions: :deletions,
+              all: :lines_of_code
+            }
+
+            type_hash[type]
+          end
+
+          it 'returns a warning when using default parameters in a PR that has larger diff than the default maximum' do
```
mokagio: Nitpick. The plugin doesn't "return" a warning, it "reports" one. Or, even if it does return the warning, that's not what we are testing here when reading `@dangerfile.status_report...`

spec/pr_size_checker_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+        end
+      end
+
+      context 'when checking a PR body size' do
+        it 'returns a warning when using default parameters in a PR that has a smaller body text length than the default minimum' do
+          allow(@plugin.github).to receive(:pr_body).and_return('pr body')
+
+          @plugin.check_pr_body
+
+          expect(@dangerfile.status_report[:warnings]).to eq ['Please provide a summary of the changes in the Pull Request description']
```
mokagio: Similar observation to the message for the diff size: This warning is correct but doesn't tell the author what they did wrong. 

What do you think of this alternative?

> The PR description appears very short, less than 10 characters long. Please provide a summary of your changes in the PR description.

spec/pr_size_checker_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+
+          expect(@dangerfile.status_report[:warnings]).to be_empty
+        end
+
+        it 'returns an error when using default parameters in a PR that has an equal pr body text length than the default minimum' do
+          allow(@plugin.github).to receive(:pr_body).and_return('some test-')
+
+          @plugin.check_pr_body
+
+          expect(@dangerfile.status_report[:warnings]).to eq ['Please provide a summary of the changes in the Pull Request description']
```
mokagio:
```suggestion
        it 'returns a warning when using default parameters in a PR that has an equal pr body text length than the default minimum' do
          allow(@plugin.github).to receive(:pr_body).and_return('some test-')

          @plugin.check_pr_body

          expect(@dangerfile.status_report[:warnings]).to eq ['Please provide a summary of the changes in the Pull Request description']
```

Or, if my nitpick about reporting vs returning resonates,

```suggestion
        it 'reports a warning when using default parameters in a PR that has an equal pr body text length than the default minimum' do
          allow(@plugin.github).to receive(:pr_body).and_return('some test-')

          @plugin.check_pr_body

          expect(@dangerfile.status_report[:warnings]).to eq ['Please provide a summary of the changes in the Pull Request description']
```

.rubocop.yml
```diff
@@ -5,6 +5,9 @@ require:
 AllCops:
   NewCops: enable
 
+Style/HashSyntax:
+  EnforcedShorthandSyntax: either
```
mokagio, line 9: My understanding is that this allows both `foo: 'bar'` and `:foo => 'bar'`, is that right?

If so, I'd personally would rather we enforced one style or the other, for consistency. (FWIW, my favorite is `foo: 'bar').

iangmaia: This is actually for checking the use of the explicit hash values -- which will also be reported for method calls with named parameters and this is why I had to use `either` 🙃 (`consistent`, which I used in another PR, failed for me here when I had a single parameter).

From [here](https://rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashSyntax):
Screenshot 2023-08-01 at 10 35 41

spec/pr_size_checker_spec.rb
iangmaia: I think in general this isn't really a must, **except** on plugins like this one that can report either warnings or errors (which is something we might extend to all plugins). I think it makes sense to check for both 👍 I'll also add a note to myself to review all other plugins to make sure they are also following the same pattern, thanks!

spec/pr_size_checker_spec.rb
iangmaia: Agree -- I believe I used "report" on another plugin; I'll also make sure to review all of them so they're consistent 👍

spec/pr_size_checker_spec.rb
iangmaia: When I read this suggestion I wondered if it wouldn't be an overkill (it's a `shared_examples` with 3 examples); but I tried it out and liked how it broke down the statements, it is more readable now. Thanks for the suggestion!

spec/pr_size_checker_spec.rb
iangmaia: I had the same thought when I first implemented something like this, but I wasn't sure about the added value of actually having separate checks for each of the combinations (given they're relatively simple and should be covered in other tests).
But I do agree that in general it is a good practice to have single-purpose, isolated tests.

I gave it a try with more shared examples to avoid the repetition, please let me know what you think: 63e01fb98823beb32df276f96612eff92b1871f7

.rubocop.yml
mokagio: Gotcha. Evidently, my understanding was wrong.

spec/pr_size_checker_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+            end
+
+            context 'when a custom error message is given and fail on error is true' do
+              include_examples 'reporting diff size custom warnings or errors', true, 'this is my custom error message'
+            end
+          end
+        end
+
+        shared_examples 'using a file selector to filter and count the changes in a diff' do |type, max_sizes|
+          context 'when using a files filter that will regard the diff as too large' do
```
mokagio, line 98: I like the verb "regard" here 😄

lib/dangermattic/plugins/pr_size_checker.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin to check the size of a Pull Request content and text body.
+  class PRSizeChecker < Plugin
+    DEFAULT_MAX_DIFF_SIZE = 500
+    DEFAULT_DIFF_SIZE_MESSAGE = "This PR is larger than #{DEFAULT_MAX_DIFF_SIZE} lines of changes. Please consider splitting it into smaller PRs for easier and faster reviews.".freeze
```
mokagio: Nitpick: the default diff message uses the default max diff size. That is appropriate but might result in inconsistent messages when the `max_size` is set but `message` isn't.

I don't consider this a blocker, but I think we could improve it by using a `Proc` instead:

```ruby
DEFAULT_DIFF_SIZE_MESSAGE = Proc.new { |size| "This PR is larger than #{size} lines of changes. Please consider splitting it into smaller PRs for easier and faster reviews." }
```

_I haven't tested the code_

mokagio (approved):

mokagio:
Thank you for taking the time to consider my suggestions.

The test rewrite came better than I expected, I think you used the shared examples well to make it DRY

lib/dangermattic/plugins/pr_size_checker.rb
iangmaia: >  the default diff message uses the default max diff size. That is appropriate but might result in inconsistent messages when the max_size is set but message isn't.

Well spotted!
Using a `Proc` works but it seemed to me a slightly more complex construct than it needed to be (specially if you want to refer to the constant from the outside of the class for some reason); I have tried to use a format instead, which also seems to work fine (929ecae79afeb6a3d60cb379a5ad1f099381fe98):

```ruby
DEFAULT_DIFF_SIZE_MESSAGE_FORMAT = "This PR is larger than %d lines of changes. Please consider splitting it into smaller PRs for easier and faster reviews.".freeze
def check_diff_size(file_selector: nil, type: :all, max_size: DEFAULT_MAX_DIFF_SIZE, message: format(DEFAULT_DIFF_SIZE_MESSAGE_FORMAT, max_size), fail_on_error: false)
```

PR #13, opened 2023-08-03
Plugin for checking the milestone in a pull request

iangmaia:
This PR adds a plugin to check the milestone assigned to a Pull Request: whether there is one assigned and whether the chosen milestone's due date is getting close.

It is based on the following Peril rules (which are essentially the same):
- https://github.com/Automattic/peril-settings/blob/master/org/pr/milestone.ts
- https://github.com/Automattic/peril-settings/blob/master/org/pr/milestone-woo.ts

This plugin provides the same functionality as the Peril rules above, but with parameters for:
- Requiring a milestone to be always set in the PR
- Customising the number of days before the milestone's due date when a warning is issued

### DIFF
--- a/lib/dangermattic/plugins/milestone_checker.rb
+++ b/lib/dangermattic/plugins/milestone_checker.rb
@@ -10,7 +10,7 @@ module Danger
       warn('PR is not assigned to a milestone.', sticky: false) if milestone.nil?
     end
 
-    # Checks if the pull request's milestone is expiring within a certain number of days.
+    # Checks if the pull request's milestone is due to finish within a certain number of days.
     #
     # @param warning_days [Integer] Number of days to warn before the milestone due date (default: DEFAULT_WARNING_DAYS).
     def check_milestone_due_date(warning_days: DEFAULT_WARNING_DAYS)
@@ -26,13 +26,13 @@ module Danger
       time_before_due_date = due_date.to_time.to_i - today.to_time.to_i
       return unless time_before_due_date <= warning_threshold
 
-      message_text = "This PR is assigned to the milestone [#{milestone['title']}](#{milestone['html_url']}) "
+      message_text = "This PR is assigned to the milestone [#{milestone['title']}](#{milestone['html_url']}). "
       message_text += if time_before_due_date.positive?
-                        "which is expiring in less than #{warning_days} days.\n"
+                        "This milestone is due in less than #{warning_days} days.\n"
                       else
-                        "which has already expired.\n"
+                        "The due date for this milestone has already passed.\n"
                       end
-      message_text += 'Please make sure to get it merged by then or assign it to a later expiring milestone.'
+      message_text += 'Please make sure to get it merged by then or assign it to a milestone with a later deadline.'
 
       warn(message_text)
     end
--- a/lib/dangermattic/plugins/milestone_checker.rb
+++ b/lib/dangermattic/plugins/milestone_checker.rb
@@ -10,10 +10,11 @@ module Danger
       return unless milestone.nil?
 
       message = 'PR is not assigned to a milestone.'
+      sticky = false
       if fail_on_error
-        failure(message, sticky: false)
+        failure(message, sticky: sticky)
       else
-        warn(message, sticky: false)
+        warn(message, sticky: sticky)
       end
     end
 
--- a/lib/dangermattic/plugins/milestone_checker.rb
+++ b/lib/dangermattic/plugins/milestone_checker.rb
@@ -5,18 +5,16 @@ module Danger
   class MilestoneChecker < Plugin
     DEFAULT_WARNING_DAYS = 5
 
+    # Checks if the pull request is assigned to a milestone.
+    def check_milestone_set
+      warn('PR is not assigned to a milestone.', sticky: false) if milestone.nil?
+    end
+
     # Checks if the pull request's milestone is expiring within a certain number of days.
     #
-    # @param needs_milestone [Boolean] Whether a milestone is required for the pull request.
-    # If true, will report a warning when there is no milestone assigned to the pull request.
     # @param warning_days [Integer] Number of days to warn before the milestone due date (default: DEFAULT_WARNING_DAYS).
-    def check_milestone_due_date(needs_milestone: true, warning_days: DEFAULT_WARNING_DAYS)
-      milestone = github.pr_json['milestone']
-
-      if milestone.nil?
-        warn('PR is not assigned to a milestone.', sticky: false) if needs_milestone
-        return
-      end
+    def check_milestone_due_date(warning_days: DEFAULT_WARNING_DAYS)
+      return if milestone.nil?
 
       milestone_due_date = milestone['due_on']
       return unless github.pr_json['state'] != 'closed' && milestone_due_date
@@ -37,6 +35,10 @@ module Danger
       message_text += 'Please make sure to get it merged by then or assign it to a later expiring milestone.'
 
       warn(message_text)
+    end
+
+    def milestone
+      github.pr_json['milestone']
     end
   end
 end
--- a/lib/dangermattic/plugins/milestone_checker.rb
+++ b/lib/dangermattic/plugins/milestone_checker.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Danger
+  # Plugin for performing checks on a milestone associated with a pull request.
+  class MilestoneChecker < Plugin
+    DEFAULT_WARNING_DAYS = 5
+
+    # Checks if the pull request's milestone is expiring within a certain number of days.
+    #
+    # @param needs_milestone [Boolean] Whether a milestone is required for the pull request.
+    # If true, will report a warning when there is no milestone assigned to the pull request.
+    # @param warning_days [Integer] Number of days to warn before the milestone due date (default: DEFAULT_WARNING_DAYS).
+    def check_milestone_due_date(needs_milestone: true, warning_days: DEFAULT_WARNING_DAYS)
+      milestone = github.pr_json['milestone']
+
+      if milestone.nil?
+        warn('PR is not assigned to a milestone.', sticky: false) if needs_milestone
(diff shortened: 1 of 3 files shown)


spec/milestone_checker_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        date_one_day_before_due = DateTime.parse('2023-06-29T23:59:01Z')
+
+        allow(@plugin.github).to receive(:pr_json).and_return(pr_json)
+        allow(DateTime).to receive(:now).and_return(date_one_day_before_due)
+
+        @plugin.check_milestone_due_date
+
+        expected_warning = ["This PR is assigned to the milestone [Release Day](https://wp.com) which is expiring in less than 5 days.\nPlease make sure to get it merged by then or assign it to a later expiring milestone."]
```
mokagio: Nitpick

```suggestion
        expected_warning = ["This PR is assigned to the milestone [Release Day](https://wp.com) which is expiring in less than 5 days.\nPlease make sure to get it merged by then or assign it to a later milestone."]
```

```suggestion
        expected_warning = ["This PR is assigned to the milestone [Release Day](https://wp.com) which is expiring in less than 5 days.\nPlease make sure to get it merged by then or assign it to a milestone expiring later."]
```

I also wonder if "expiring" is the right verb to use here 🤔 . What to you think of "due to finish" or "with deadline"?

spec/milestone_checker_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        allow(@plugin.github).to receive(:pr_json).and_return(pr_json)
+
+        @plugin.check_milestone_due_date
+
+        expect(@dangerfile.status_report[:warnings]).to be_empty
+      end
+
+      it 'does nothing when a PR has a milestone but already closed' do
```
mokagio: Maybe out of scope for this PR that ports the existing rule, but I wonder if it would be best to emit a warning anyways, to let the author know they might have picked the wrong milestone, or that an old PR that just received changed needs its milestone updated.

spec/milestone_checker_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      it 'warns when a PR without a milestone' do
+        allow(@plugin.github).to receive(:pr_json).and_return({})
+
+        @plugin.check_milestone_due_date
+
+        expected_warning = ['PR is not assigned to a milestone.']
+        expect(@dangerfile.status_report[:warnings]).to eq expected_warning
+      end
+
+      it 'does nothing when a PR does not have a milestone but this isnt required' do
```
mokagio: Nitpick

```suggestion
      it "does nothing when a PR does not have a milestone but this isn't required" do
```

---

Having said that, is this a parameter and behavior we need? 

I would assume that if we were to setup a repo that doesn't check for milestones, e.g. any of the library repos, we'd merely not add the check to the `Dangerfile`.

mokagio: I've re-read the diff and I think I get the intention: Use the same method to 1) check PRs have a milestone 2) check the milestone, if any, is still open.

Personally, I see those as two different checks, which could share implementation details under the hood.

- Ensure the PR has a milestone (similar to ensure the PR has a label)
- Ensure the milestone is still open (which implicitly assumes the PR should have a milestone and will fail/warn otherwise)

Keen to hear what you think.

iangmaia: > I've re-read the diff and I think I get the intention: Use the same method to 1) check PRs have a milestone 2) check the milestone, if any, is still open.

Exactly; this is something inherited from the original Peril rule, just without the optionality on the milestone existence check.
I do agree it can be split into two separate checks, but now I'm not sure if the second check should also warn if the PR doesn't have a milestone. We could remove the check from the second, so that the caller can combine both calls if they want, having the freedom to have the milestone check or not (that was what I initially intended with the parameter):
```ruby
# Dangerfile

# will check if there's a milestone
milestone_checker.check_milestone_set

# will check if a milestone is close do its due date, but just return when there's no milestone set
milestone_checker.check_milestone_due_date
```

Having both calls is maybe what many repos will want, and we can have this behaviour in a common `Dangerfile` together with other common checks, so that other repos can just import it.

What do you think?

spec/milestone_checker_spec.rb
iangmaia: Good point! Perhaps it can be a different warning message altogether.

spec/milestone_checker_spec.rb
iangmaia: I had the same question with expiring / closing or just say its due date already passed and related. I think "expiring" is not necessarily wrong but it sounds maybe too informal. I'll try rewording it.

spec/milestone_checker_spec.rb
mokagio: > We could remove the check from the second, so that the caller can combine both calls if they want, having the freedom to have the milestone check or not (that was what I initially intended with the parameter)

🤔 

So far, my mental model has been "if you check that the milestone has a due date, then there should be an error if the milestone is not there". But I can see how this would work in a conditional mode such as "if there's a milestone, verify its due date is within the accepted range".

🤔 

From that angle, the current implementation gives more latitude.

What do you think of putting it all together, like this

```ruby
# Dangerfile

# will check if there's a milestone
milestone_checker.check_milestone_set

# will check if a milestone is close do its due date, but just return when there's no milestone set
milestone_checker.check_milestone_due_date

# will check if a milestone is close do its due date, and report an error if there's no milestone set
milestone_checker.check_milestone_due_date(if_no_milestone: :error)

# will check if a milestone is close do its due date, and report a warning if there's no milestone set
milestone_checker.check_milestone_due_date(if_no_milestone: :warn)
```

iangmaia: I like the idea of the parameter like that. It makes the intention clearer plus providing the option to have warning/error/nothing when there's no milestone. I've pushed:
- Breaking down into two methods: 6d6a8e261222134cf7dfbd184eae67c698c774a7
- Wording update: 31f6aafa049ef8d6b1b8b557d8815165a28cf1db
- The new parameter: fbe857c7efa47c1a431da4084591a5941e6a96db

lib/dangermattic/plugins/milestone_checker.rb
```diff
@@ -0,0 +1,58 @@
...
+    # Checks if the pull request is assigned to a milestone.
+    def check_milestone_set(fail_on_error: false)
+      return unless milestone.nil?
+
+      message = 'PR is not assigned to a milestone.'
+      if fail_on_error
+        failure(message, sticky: false)
+      else
+        warn(message, sticky: false)
+      end
```
mokagio: Nitpick:

```suggestion
      sticky = false
      if fail_on_error
        failure(message, sticky: sticky)
      else
        warn(message, sticky: sticky)
      end
```

lib/dangermattic/plugins/milestone_checker.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    # Checks if the pull request's milestone is due to finish within a certain number of days.
+    #
+    # @param warning_days [Integer] Number of days to warn before the milestone due date (default: DEFAULT_WARNING_DAYS).
+    # @param if_no_milestone [Symbol] Action to take if the pull request is not assigned to a milestone. Possible values:
+    #                 - :warn (default): Reports a warning.
+    #                 - :error: Reports an error.
+    #                 - :none or nil: Takes no action.
+    def check_milestone_due_date(warning_days: DEFAULT_WARNING_DAYS, if_no_milestone: :warn)
```
mokagio: Nitpick: Given that the behavior now allows for reporting an error, too, `warning_days` is not 100% aligned as a name.

What do you think of `days_threshold` or `days_cutoff`? I don't love either, but I'm trying to describe the "threshold of days left below which to ~warn~ report".

spec/milestone_checker_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+          expected_error = ['PR is not assigned to a milestone.']
+          expect(@dangerfile.status_report[:errors]).to eq expected_error
+        end
+
+        it "does nothing when asked to do so when a PR doesn't have a milestone set" do
+          allow(@plugin.github).to receive(:pr_json).and_return({})
+
+          @plugin.check_milestone_due_date(if_no_milestone: :none)
+
+          expect(@dangerfile.status_report[:errors]).to be_empty
```
mokagio, line 222: I think this might be a false positive test. 

The current implementation is calls

```ruby
        check_milestone_set(fail_on_error: if_no_milestone == :error)
```

meaning that in this case it'll report a warning.

mokagio: Confirmed locally:

image

mokagio: This instance reinforces my belief that we would benefit from consistently checking both errors and warnings.

Maybe we could write a custom expectation or just a util method in `spec_helper` 🤔 Anyways, that's something for another PR.

mokagio (commented):
Just a minor thing in the error / warn / nothing behavior

spec/milestone_checker_spec.rb
iangmaia: Nice catch! When I introduced the error/warning, the tests also should have been updated to check for both.
I've fixed this here e51cb9a83bf03ca6bb095aae5f42c7129414165c and here a6359917bac2346cd060364a2b50dab75232f6cb.

While I'm at it, I'm going to create a PR on top of this one to use a new expectation util method 👍

mokagio (approved):
Thank you for addressing my suggestions.

lib/dangermattic/plugins/milestone_checker.rb
iangmaia: Indeed; I've tried `days_before_due: DEFAULT_DAYS_THRESHOLD` (cc67ee757c47d5eaa56eaa5c0949dc0d61be3bdb). I think it is kinda hard to have a really concise name, the parameter description should help 🤷

PR #14, opened 2023-08-09
Add custom RSpec Matchers for checking Danger errors and warnings

iangmaia:
Based on my recent discussions with @mokagio on #13, this PR builds on top of #13 and adds custom matchers to be used with expectations, so that instead of:

```ruby
# expect empty warning/error reports
expect(@dangerfile.status_report[:warnings]).to be_empty
expect(@dangerfile.status_report[:errors]).to be_empty

# expect no warnings reported and the specified errors
expected_errors = ...
expect(@dangerfile.status_report[:warnings]).to be_empty
expect(@dangerfile.status_report[:errors]).to eq expected_errors

# expect no errors reported and the specified warnings
expected_warnings = ...
expect(@dangerfile.status_report[:errors]).to be_empty
expect(@dangerfile.status_report[:warnings]).to eq expected_warnings
```

We can do:
```ruby
# expect empty warning/error reports
expect(@dangerfile).to contain_empty_report

# expect no warnings reported and the specified errors
expected_errors = ...
expect(@dangerfile).to contain_errors(expected_errors)

# expect no errors reported and the specified warnings
expected_warnings = ...
expect(@dangerfile).to contain_warnings(expected_warnings)
```

### DIFF
--- a/spec/milestone_checker_spec.rb
+++ b/spec/milestone_checker_spec.rb
@@ -22,7 +22,7 @@ module Danger
             @plugin.check_milestone_set
 
             expected_warning = ['PR is not assigned to a milestone.']
-            expect(@dangerfile).to contain_warnings(expected_warning)
+            expect(@dangerfile).to report_warnings(expected_warning)
           end
 
           it "reports an error when a PR doesn't have a milestone set" do
@@ -31,7 +31,7 @@ module Danger
             @plugin.check_milestone_set(fail_on_error: true)
 
             expected_error = ['PR is not assigned to a milestone.']
-            expect(@dangerfile).to contain_errors(expected_error)
+            expect(@dangerfile).to report_errors(expected_error)
           end
         end
 
@@ -47,7 +47,7 @@ module Danger
 
             @plugin.check_milestone_set
 
-            expect(@dangerfile).to contain_empty_report
+            expect(@dangerfile).to do_not_report
           end
 
           it 'does nothing when an error is expected but the PR has a milestone set' do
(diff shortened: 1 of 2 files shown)


mokagio (commented):
Glad to see you liked my suggestion 😄 

One nitpick. "expect dangerfile to contain..." made me think that the warnings/errors are in the `Dangerfile`, which is not exact.

What do you think of either `expect(@dangerfile.status_report).to contain_warnings` or `expect(@dangerfile).to report_warnings`?

iangmaia:
> Glad to see you liked my suggestion 😄
> 
> One nitpick. "expect dangerfile to contain..." made me think that the warnings/errors are in the `Dangerfile`, which is not exact.
> 
> What do you think of either `expect(@dangerfile.status_report).to contain_warnings` or `expect(@dangerfile).to report_warnings`?

I struggled a bit with it -- my initial code used `expect(@dangerfile.status_report).to contain_warnings`, but to me seemed too verbose to use repeatedly while we can access the report from `@dangerfile` in the matcher itself.

`expect(@dangerfile).to report_warnings` is a good suggestion. At the same time "expect dangerfile to report warnings" falls into a similar problem as "expect dangerfile to contain warnings", as in not exactly reflecting what is happening; though it is not wrong either. Still I think that reads better and is more accurate, so I'll use it, thanks! 👍 

Update: done here 4ae61538ea29fcd9e5c8566411d0404dbc82fc09

mokagio (approved):
I'm always glad when the biggest issues we have are hairsplitting naming decisions 😄

PR #15, opened 2023-08-15
Add plugin for checking Tracks usage

iangmaia:
This plugin ports the [check-tracks.ts](https://github.com/Automattic/peril-settings/blob/master/org/pr/check-tracks.ts) Peril rule, which is currently running only on Woo iOS/Android.
The plugin works as following: if there is a change in one of the specified files OR if there is a certain expression found within the diff content, the plugin will report a message reminding the author/reviewer to validate the events and verify the internal tracks spreadsheet. The message also mentions a **Tracks** label, which could be implemented as a check instead using https://github.com/Automattic/dangermattic/pull/7 in the near future.

I've created this PR on top of the https://github.com/Automattic/dangermattic/pull/10 PR so that I could also use the shared plugin `git_utils`, used here to check each diff line against a matching expression.


spec/check_tracks_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+      describe '#check_tracks_changes' do
+        before do
+          allow(@plugin.git).to receive(:modified_files).and_return([])
+          allow(@plugin.git).to receive(:added_files).and_return([])
+          allow(@plugin.git).to receive(:deleted_files).and_return([])
+
+          stub_const('GitDiffStruct', Struct.new(:type, :path, :patch))
+        end
+
+        context 'when checking changes in track related files' do
```
mokagio:
```suggestion
        context 'when checking changes in Tracks related files' do
```

spec/check_tracks_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+          it 'reports a message with instructions for review when there are changes in tracks files' do
```
mokagio:
```suggestion
          it 'reports a message with instructions for review when there are changes in Tracks related files' do
```

spec/check_tracks_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+            allow(@plugin.git_utils).to receive(:all_changed_files).and_return(['Test.kt', 'LoginAnalyticsTracker.kt', 'Test.java'])
+
+            @plugin.check_tracks_changes
+
+            expect(@dangerfile.status_report[:messages]).to eq [CheckTracks::TRACKS_PR_INSTRUCTIONS]
+          end
+
+          it 'reports a message with instructions for review when there are changes in tracks files using a custom file list' do
```
mokagio:
```suggestion
          it 'reports a message with instructions for review when there are changes in Tracks related files using a custom file list' do
```

spec/check_tracks_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+            allow(@plugin.git_utils).to receive(:all_changed_files).and_return(['MyClass.swift', 'MyClass1.swift', 'MyClass2.swift'])
+
+            @plugin.check_tracks_changes(tracks_files: ['MyClass1.swift'])
+
+            expect(@dangerfile.status_report[:messages]).to eq [CheckTracks::TRACKS_PR_INSTRUCTIONS]
+          end
+
+          it 'does nothing when there are no changes in tracks files' do
```
mokagio:
```suggestion
          it 'does nothing when there are no changes in Tracks related files' do
```

spec/check_tracks_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+            modified_files = ['MyClass.swift']
+            allow(@plugin.git_utils).to receive(:all_changed_files).and_return(modified_files)
+            allow(@plugin.git_utils).to receive(:match_diff_lines_in_files).with(files: modified_files, line_matcher: kind_of(Proc), change_type: nil).and_return([])
+
+            @plugin.check_tracks_changes
+
+            expect(@dangerfile.status_report[:messages]).to be_empty
+          end
+
+          it 'does nothing when there are no changes in tracks files using a custom file list' do
```
mokagio:
```suggestion
          it 'does nothing when there are no changes in Tracks related files using a custom file list' do
```

spec/check_tracks_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+            allow(@plugin.git_utils).to receive(:all_changed_files).and_return(modified_files)
+            allow(@plugin.git_utils).to receive(:match_diff_lines_in_files).with(files: modified_files, line_matcher: kind_of(Proc), change_type: nil).and_return([])
+
+            @plugin.check_tracks_changes(tracks_files: ['MyClass1.swift'])
+
+            expect(@dangerfile.status_report[:messages]).to be_empty
+          end
+        end
+
+        context 'when checking track related changes within a diff' do
```
mokagio:
```suggestion
        context 'when checking Tracks related changes within a diff' do
```

lib/dangermattic/plugins/check_tracks.rb
```diff
@@ -0,0 +1,56 @@
...
+    DEFAULT_TRACKS_USE_MATCHERS = [
+      /AnalyticsTracker\.track/
+    ].freeze
+
+    TRACKS_PR_INSTRUCTIONS = <<~MESSAGE
+      This PR contains changes to Tracks-related logic. Please ensure the following are completed:
+      **PR Author**
+      - The PR must be assigned the **Tracks** label
+      **PR Reviewer**
+      - The tracks events must be validated in the Tracks system.
```
mokagio: I feel like one of the "tracks" is redundant in the message. There's enough context to understand that the event is a Tracks event

```suggestion
      - The events must be validated in the Tracks system.
```

lib/dangermattic/plugins/check_tracks.rb
```diff
@@ -0,0 +1,56 @@
...
+      - Verify the internal tracks spreadsheet has also been updated.
```
mokagio:
```suggestion
      - Verify the internal Tracks spreadsheet has also been updated.
```

lib/dangermattic/plugins/check_tracks.rb
```diff
@@ -0,0 +1,56 @@
...
+    MESSAGE
+
+    # Checks the PR diff for changes in tracks-related files and provides instructions if changes are detected
```
mokagio:
```suggestion
    # Checks the PR diff for changes in Tracks-related files and provides instructions if changes are detected
```

lib/dangermattic/plugins/check_tracks.rb
```diff
@@ -0,0 +1,56 @@
...
+    #
+    # @param tracks_files [Array<String>] List of tracks-related file names to check (default: DEFAULT_TRACKS_FILES)
```
mokagio:
```suggestion
    # @param tracks_files [Array<String>] List of Tracks-related file names to check (default: DEFAULT_TRACKS_FILES)
```

lib/dangermattic/plugins/common/git_utils.rb
```diff
@@ -21,16 +21,17 @@ class GitUtils < Plugin
     def check_added_diff_lines(file_selector:, line_matcher:, message:, fail_on_error: false)
       modified_files = added_and_modified_files.select(&file_selector)
 
-      modified_files.each do |file|
-        diff = danger.git.diff_for_file(file)
-
-        diff.patch.each_line do |line|
-          next unless change_type(diff_line: line) == :added
-          next unless line_matcher.call(line)
+      matches = match_diff_lines_in_files(
```
mokagio: What do you think of calling the method `matching_lines_in_diff` or `matching_lines_in_diff_files`? The verb `match` doesn't make me thing of a method that would necessarily return something

```ruby
      matches = matching_lines_in_diff(files: f, line_matcher: l, change_type: :added)
```

lib/dangermattic/plugins/check_tracks.rb
```diff
@@ -0,0 +1,56 @@
...
+    DEFAULT_TRACKS_FILES = [
+      'AnalyticsTracker.kt',
+      'AnalyticsEvent.kt',
+      'LoginAnalyticsTracker.kt',
+      'WooAnalyticsStat.swift'
+    ].freeze
+
+    DEFAULT_TRACKS_USE_MATCHERS = [
+      /AnalyticsTracker\.track/
+    ].freeze
```
mokagio, line 27: How do you feel about these default values? 

I think they were necessarily hardcoded in the Peril implementation because all rules had to be specified in the Peril settings repo. In the context of a Danger plugin, I fell like these would be useful defaults only for WooCommerce (see `WooAnalyticsStat.swift` for example!). 

Removing the defaults would make setting up the plugin more cumbersome. At the same time, I like the idea of having the configuration for what is Tracks-related live in the repo for which it applies.

mokagio (commented):
If you haven't done it already, do you think it would be worth asking the Woo team whether they find value in this check? It would be a shame to do all the work to set it up only to backfire on us. Intuitively, this seems very useful, but I'm not the one working daily in the codebase and tweaking Tracks.

I left a bunch of nitpicky wording suggestions based on the use of a compound adjective and the fact that Tracks is a noun. I decided to open a PR for them, to underscore those a things that _I_ find valuable but not necessary for this PR to be "good". See https://github.com/Automattic/dangermattic/pull/17

Curious to hear your thoughts on the default use. 

Finally, I feel these changes would benefit from the new matcher from #14 . I guess it was not available at the time of creating the PR.

lib/dangermattic/plugins/check_tracks.rb
iangmaia: (gathering replies from the [initial comment](https://github.com/Automattic/dangermattic/pull/15#pullrequestreview-1592553671) as it is kinda related to this one)

> If you haven't done it already, do you think it would be worth asking the Woo team whether they find value in this check?

I haven't. Since this is a very process-specific check, that's a good idea.

> Intuitively, this seems very useful

That's what I thought, but that was also an assumption based on what I saw in Peril. I'll write to them anyway.

One thing that I thought was that this check mostly builds on top of generic code and it could be a template for similar cases; so, even if this isn't applicable right now, this plugin could be easily repurposed to a generic case (as in: "warn/message/error" on files/patterns inside the diff).

> How do you feel about these default values?

I had mixed feelings about it, since it is very team-specific (Woo iOS/Android) and these plugin checks should be as general purpose as possible. So I thought it could be a repo/team specific check; alternatively, in this case, perhaps it would be even worth it moving this plugin code to Woo's Dangerfile.
And since this was running only on Woo, I'm not sure if it would be useful for other repos. Maybe for checking completely different things other than Tracks, with a different message?...

> In the context of a Danger plugin, I fell like these would be useful defaults only for WooCommerce

I think this is what is the key here: context of a Danger plugin vs. more specific checks.
I am not sure yet how to deal with those checks that are very team specific, or even to a couple of teams but not really general purpose.
In principle, my idea for these cases was to use shared specialised `Dangerfile`s that could be imported by the projects (e.g. possibly have a `woo/Dangerfile` inside `dangermattic`), but I haven't reached the point yet to decide to go for this approach instead of using more specific plugins.
For now, I just ported what was in Peril to then decide how to move the code around once we start integrating these checks into the projects.

lib/dangermattic/plugins/common/git_utils.rb
iangmaia: Agreed!

iangmaia:
@mokagio I've added handling of `messages` reports to the custom expectation matcher from #14 (as we only had errors/warnings reports so far) so that I could use it in this PR, see 1098d6a35bf2a61e1ff30ac213417172c4bd1bd4.

lib/dangermattic/plugins/check_tracks.rb
mokagio: > I think this is what is the key here: context of a Danger plugin vs. more specific checks.

🤔 it's tough to know what to do a priori and I don't want my questions/suggestions to get us bogged down implementing abstractions that make sense but that we don't need to use yet.

Given "In the context of a Danger plugin, I fell like these would be useful defaults only for WooCommerce," I guess as long as the plugin/check is only used in Woo the setup does the job. We can iterated later when using it on another app.

mokagio (approved):
Code looking good, thank you for the refinements. Let's see how it plays up IRL 😄

PR #18, opened 2023-08-29
Publishing Gem to RubyGems

iangmaia:
I have made a couple changes in this Pull Request:

- Firstly, I've updated the name of the Gem to `danger-dangermattic`, to align it with the [convention for Danger plugins](https://danger.systems/guides/creating_your_first_plugin#creating-the-plugin)
- In addition to the renaming, I've also taken steps to prepare for the publication of the Gem on RubyGems.
- As part of the process, I've updated RuboCop, which required some modifications to a test in `pr_size_checker_spec.rb`.


.buildkite/pipeline.yml
```diff
@@ -36,3 +36,15 @@ steps:
...
+  # Push to RubyGems
+  #################
+  - label: ":rubygems: Publish to RubyGems"
+    key: "gem-push"
+    if: build.tag != null
+    depends_on:
+     - test
+     - rubocop
+    command: .buildkite/gem-push.sh
+    plugins: *common_plugins
```
AliSoftware: You'll need to have the docker plugin be told to forward the `RUBYGEMS_API_KEY` environment variable to its container too, otherwise when the `gem push` command will be run from within the docker container it will not know it.

You can do that like this when providing the plugin definition to `plugins:` like this
```yaml
    - docker#v3.8.0:
        image: &ruby_version "ruby:2.7.4"
        propagate-environment: true
        environment:
          - "RUBYGEMS_API_KEY"
```

There's a question weather it's ok to declare that `environment` to be propagated… as part of the `&common_plugin` YAML anchor — which means that we'd propagate this env var in all steps that use `plugins: *common_plugins`, including the ones that might not need that env var — or to create separate YAML anchors depending on which env vars each step that are using the docker plugin need to expose (which makes the configuration and the YAML a bit more cumbersome, so it is really worth it?)

.buildkite/gem-push.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -eu
+
+GEM_NAME="danger-dangermattic"
+
+echo "--- :hammer: Build Gemspec"
+gem build "$GEM_NAME.gemspec" -o "$GEM_NAME.gem"
```
AliSoftware, line 6: The `.gemspec` file is currently still named `dangermattic.gemspec` not `danger-dangermattic.gemspec`. We might want to rename it

dangermattic.gemspec
```diff
@@ -7,7 +7,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'dangermattic/gem_version'
 
 Gem::Specification.new do |spec|
-  spec.name          = 'dangermattic'
+  spec.name          = 'danger-dangermattic'
```
AliSoftware, line 10: I'm slightly surprised that rubocop is ok with the name of the `.gemspec` file differs from the value of the `spec.name` here…
We'll have to rename the file to `danger-dangermattic.gemspec` anyway if we want the logic in `gem-push.sh` relying on `$GEM_NAME` to work as expected.

spec/pr_size_checker_spec.rb
```diff
@@ -16,9 +16,7 @@ module Danger
 
       context 'when checking a PR diff size' do
         before do
-          allow(@plugin.git).to receive(:added_files).and_return([])
-          allow(@plugin.git).to receive(:modified_files).and_return([])
-          allow(@plugin.git).to receive(:deleted_files).and_return([])
+          allow(@plugin.git).to receive_messages(added_files: [], modified_files: [], deleted_files: [])
```
AliSoftware, line 19: TIL `receive_messages` to group a set of mockings in one call. Nice!

AliSoftware (changes requested):

.buildkite/pipeline.yml
iangmaia: I wasn't aware of this but it makes sense, thanks.
I wonder if I could somehow reuse the `common_plugins` container with the right image and somehow just "override" the environment (I quickly searched in the Buildkite docs and couldn't find it).

So for now I just declared the container again only for that step, propagating the environment: 4ee5a0760b080b82f94cd3e741d7bde836de65e9

AliSoftware: Yeah it's not really possible sadly, at least not directly in YAML (we'd have to do some custom pre-processing of the YAML pipeline before piping it to `buildkite-agent pipeline upload` if we wanted to support that — see paaHJt-50A-p2).

What you could do though and that I'd recommend (mostly so that we keep the constants and list of plugin versions used throughout the pipeline all grouped at the top for easier future maintenance) is to [create YAML anchors for the versions and images (and any other entries or strings you're repeating)](https://mobile.blog/2021/09/09/clean-config-files-with-yaml-anchors/), even if those anchors don't encapsulate the whole YAML node describing the whole plugin.

Something like this, for example:
```yaml
common_params:
  - &docker_plugin     "docker#v5.7.0"
  - &docker_ruby_image "public.ecr.aws/docker/library/ruby:3.2.2"
  - &common_plugins
    - *docker_plugin :
        image: *docker_ruby_image

steps:
  - …
  - label: "🧹 Lint (Rubocop)"
    key: rubocop
    command: …
    plugins: *common_plugins
  - …
  - label: ":rubygems: Publish to RubyGems"
    key: "gem-push"
    if: build.tag != null
    depends_on:
     - test
     - rubocop
    command: .buildkite/gem-push.sh
    plugins:
      - *docker_plugin :
          image: *docker_ruby_image
          propagate-environment: true
          environment:
            - "RUBYGEMS_API_KEY"
```
The only gotcha to know about is that if you want to use a string anchor as a dictionary key (like `*docker_plugin` in the example above to use the `docker_plugin` YAML anchor pointing to the `docker#v5.7.0` string), you need to be sure to have a space between the anchor name and the `:` — i.e. `*docker_plugin :` not just `*docker_plugin:`.

iangmaia: > to [create YAML anchors for the versions and images (and any other entries or strings you're repeating)](https://mobile.blog/2021/09/09/clean-config-files-with-yaml-anchors/)

Thanks for the suggestions -- nice article, by the way 🙂

iangmaia: Updated here: ~337c46337acca77527f6ca08224f2fa3b0a8b13c~ 43debc1af5de8ccbc25abed59234f98f4a25c2dd

AliSoftware (approved):

.buildkite/pipeline.yml
AliSoftware: I guess you meant 43debc1af5de8ccbc25abed59234f98f4a25c2dd 😛

PR #21, opened 2023-09-08
Rename `do_not_report` to `not_report`

mokagio:
"Expect x to not report" reads better than "Expect x to do not report," IMHO. What do you think @iangmaia ? If okay, I'll wait for the recent improvements in the Danger matchers to land on `trunk` then adjust this PR to account for them.

### DIFF
--- a/spec/milestone_checker_spec.rb
+++ b/spec/milestone_checker_spec.rb
@@ -47,7 +47,7 @@ module Danger
 
             @plugin.check_milestone_set
 
-            expect(@dangerfile).to do_not_report
+            expect(@dangerfile).to not_report
           end
 
           it 'does nothing when an error is expected but the PR has a milestone set' do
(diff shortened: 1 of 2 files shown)


iangmaia:
> "Expect x to not report" reads better than "Expect x to do not report," IMHO. What do you think @iangmaia ? If okay, I'll wait for the recent improvements in the Danger matchers to land on `trunk` then adjust this PR to account for them.

Though `not report` this crossed my mind in the beginning, I discarded it for some reason. Maybe due to the "@dangerfile" parameter not exactly matching it... but looking at it now, I think it reads much better, thanks!

I think you can merge it straight away -- there are some additional changes related to the matchers in other PRs (like checking for message reports here d952dbf3f8ad2afd6e43ced840854b0dfb74215d) but most of it is on `trunk` already. Then the pending PRs that use this matcher can be adapted when they're rebased.

mokagio:
@iangmaia thanks for your comment.

> I think you can merge it straight away

In that case, I'll mark this as ready for review

iangmaia (approved):
👍

PR #22, opened 2023-09-12
Add RuboCop Danger plugin

iangmaia:
Adding `danger-rubocop` so it becomes part of the "Dangermattic package", so that we can use it as needed in client repos without the need for adding it as a separate dependency.


iangmaia:
Gentle reminder @mokagio 🙂

mokagio (approved):
Too easy! Thanks

PR #25, opened 2023-11-03
Codefence both file and lockfile names in `manifest_pr_checker` warn

mokagio:
I got a `Podfile.lock` warning in https://github.com/wordpress-mobile/WordPress-iOS/pull/21961:

image

Just a personal opinion that those file names look better rendered as `<code>`. Also, it would look consistent with the codefenced command suggestion.

image

If this doesn't resonate, feel free to discard.


iangmaia (approved):
Looks nicer indeed 👍

PR #26, opened 2023-11-03
Prevent Danger crash due to nil Diff Stats

iangmaia:
[This WordPress Android PR](https://github.com/wordpress-mobile/WordPress-Android/pull/19488) had an interesting crash: it happened internally in Danger when calling the `git` function `info_for_file` ([see code](https://github.com/danger/danger/blob/d9fff7cbed6b0a4bfb6465709d0352e700bdd00e/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb#L138)). The GitHub API might return nil for the stats object.

The present PR fixes the issue by avoiding the `info_for_file` call when the `stats` object is `nil`.

## How to test

1. Checkout the aforementioned WP Android branch
2. Change the Gemfile to use this branch:
`gem 'danger-dangermattic', git: 'https://github.com/Automattic/dangermattic', branch: 'fix/prevent-info-for-file-crash`
3. Run `bundle update danger-dangermattic`
4. Run danger locally but pointing to the PR above:
```
DANGER_GITHUB_API_TOKEN=<token> bundle exec danger pr https://github.com/wordpress-mobile/WordPress-Android/pull/19488
```

### DIFF
--- a/lib/dangermattic/plugins/pr_size_checker.rb
+++ b/lib/dangermattic/plugins/pr_size_checker.rb
@@ -85,7 +85,13 @@ module Danger
       return danger.git.insertions unless file_selector
 
       filtered_files = git_utils.all_changed_files.select(&file_selector)
-      filtered_files.sum { |file| danger.git.info_for_file(file)[:insertions].to_i }
+
+      filtered_files.sum do |file|
+        # stats for a file in the GitHub API might be nil, making `info_for_file()` crash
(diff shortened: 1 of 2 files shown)


AliSoftware (approved):
I have NOT tested / followed the testing steps of this PR, but approving to unblock since the logic from reading the PR's diff looks good to me.

PR #27, opened 2023-11-07
Improve screenshot matching in the PR description

iangmaia:
## What
The plugin wasn't detecting the screenshots on [this PR](https://github.com/woocommerce/woocommerce-ios/pull/11097), so I tweaked the Regex a bit more, which should now better cover the `` HTML tag in the PR description.

## How to test

1. Checkout the branch from the aforementioned PR
2. Change the `Gemfile` to use the present branch:
`gem 'danger-dangermattic', git: 'https://github.com/Automattic/dangermattic', branch: 'fix/screenshots-pr-check`
3. Run `bundle install`
4. Run danger locally but pointing to the PR above:
```
DANGER_GITHUB_API_TOKEN=<token> bundle exec danger pr https://github.com/woocommerce/woocommerce-ios/pull/11097
```

### DIFF
--- a/lib/dangermattic/plugins/view_changes_need_screenshots.rb
+++ b/lib/dangermattic/plugins/view_changes_need_screenshots.rb
@@ -20,7 +20,7 @@ module Danger
     IMAGE_IN_PR_BODY_PATTERNS = [
       %r{https?://\S*\.(gif|jpg|jpeg|png|svg)},
       /!\[(.*?)\]\((.*?)\)/,
-      /<img\s+[^\>]*src\s*=\s*[^\>]*>/
+      /<img\s+[^>]*src\s*=\s*[^>]*>/
     ].freeze
 
     # Checks if view files have been modified and if a screenshot is included in the pull request body,
--- a/lib/dangermattic/plugins/view_changes_need_screenshots.rb
+++ b/lib/dangermattic/plugins/view_changes_need_screenshots.rb
@@ -17,6 +17,12 @@ module Danger
     VIEW_EXTENSIONS_IOS = /(View|Button)\.(swift|m)$|\.xib$|\.storyboard$/
     VIEW_EXTENSIONS_ANDROID = /(?i)(View|Button)\.(java|kt|xml)$/
 
+    IMAGE_IN_PR_BODY_PATTERNS = [
+      %r{https?://\S*\.(gif|jpg|jpeg|png|svg)},
+      /!\[(.*?)\]\((.*?)\)/,
+      /<img\s+[^\>]*src\s*=\s*[^\>]*>/
+    ].freeze
+
     # Checks if view files have been modified and if a screenshot is included in the pull request body,
     # displaying a warning if view files have been modified but no screenshot is included.
     #
@@ -26,9 +32,9 @@ module Danger
         VIEW_EXTENSIONS_IOS =~ file || VIEW_EXTENSIONS_ANDROID =~ file
       end
 
-      pr_has_screenshots = github.pr_body =~ %r{https?://\S*\.(gif|jpg|jpeg|png|svg)}
-      pr_has_screenshots ||= github.pr_body =~ /!\[(.*?)\]\((.*?)\)/
-      pr_has_screenshots ||= github.pr_body =~ /<img\s[^>]*\ssrc=[^>]*>/
+      pr_has_screenshots = IMAGE_IN_PR_BODY_PATTERNS.any? do |pattern|
+        github.pr_body =~ pattern
+      end
 
       warning = 'View files have been modified, but no screenshot is included in the pull request. ' \
                 'Consider adding some for clarity.'
--- a/spec/view_changes_need_screenshots_spec.rb
+++ b/spec/view_changes_need_screenshots_spec.rb
@@ -39,9 +39,18 @@ module Danger
         expect(@dangerfile.status_report[:warnings]).to be_empty
       end
 
-      it 'does nothing when a PR with view code changes has a screenshot defined with a html tag' do
+      it 'does nothing when a PR with view code changes has a screenshot defined with a html tag with different attributes before src' do
         allow(@plugin.github).to receive(:pr_body)
-          .and_return("see screenshot:\n<img width=300 src=\"https://github.com/bloom/DayOne-Apple/assets/4780/1f9e01a8-c63d-41d4-9ac8-fa9a5182ab55\"> body body")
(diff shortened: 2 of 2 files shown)


jkmassel (commented):
As a rule, parsing HTML with regex can be dicey (see https://blog.codinghorror.com/parsing-html-the-cthulhu-way/, et al)

I wonder if we could do something easier to work with, like:

```
<!-- DANGERMATTIC_IMAGE_1 -->
<img [...whatever...] />
<!-- END_OF_DANGERMATTIC_IMAGE_1 -->
```

This can avoid a lot of potential edge cases because we can just replace everything between the comment blocks (which IINM can technically be represented by a grammar in a way that HTML can't). 

WDYT?

AliSoftware:
> I wonder if we could do something easier to work with, like:
> 
> 
> 
> ```
> 
> 
> 
> 
> 
> 
> 
> ```

This would require the authors of PRs to add such long markers fences around all their image tags manually every time when they write their PR description.

Personally I feel it would likely become more annoying to PR authors and detrimental to the rule rather than making it perceived as useful; especially since the PR authors would have to remember about this and about the exact syntax/spelling of the fence magic comment.

And we can't put that comment in the PR template to remind them how to use it because doing so would trick the Danger rule into matching the occurrence from the template (many people don't remove HTML comments from the template when filling in their PR description)… and thus completely defeat the purpose of the rule 😅

lib/dangermattic/plugins/view_changes_need_screenshots.rb
```diff
@@ -17,6 +17,12 @@ class ViewChangesNeedScreenshots < Plugin
     VIEW_EXTENSIONS_IOS = /(View|Button)\.(swift|m)$|\.xib$|\.storyboard$/
     VIEW_EXTENSIONS_ANDROID = /(?i)(View|Button)\.(java|kt|xml)$/
 
+    IMAGE_IN_PR_BODY_PATTERNS = [
+      %r{https?://\S*\.(gif|jpg|jpeg|png|svg)},
+      /!\[(.*?)\]\((.*?)\)/,
+      /<img\s+[^\>]*src\s*=\s*[^\>]*>/
```
AliSoftware: I don't think it's necessary to escape the `>` inside the character class sub pattern, as it's not a special character interpreted other than a literal by Regexp… is it?

```suggestion
      /<img\s+[^>]*src\s*=\s*[^>]*>/
```

iangmaia:
> As a rule, parsing HTML with regex can be dicey (see https://blog.codinghorror.com/parsing-html-the-cthulhu-way/, et al)

A classic! 🙂 While it's hard to disagree as a principle, the scope of the problem here is very narrow; it is more like "is there an img tag at all here" than actually parsing.

In fact, in the initial version we shipped there was no attempt to detect a `` tag and only image URLs...
...but it turns out that, when you drag and drop an image to Github, it generates a URL without an image extension.
Github initially presents the image using markdown, but it's a common practice for developers to change it to a `` tag to specify the `width` (when developing app features on previous projects, I did exactly the same).

> I wonder if we could do something easier to work with, like:
> 
> ```
> 
> 
> 
> ```

I agree with @AliSoftware on this one — I think the friction to have the checks running should be minimal, and it should be as unintrusive as possible in developers' routines.

As an alternative, if the need arises also for other cases, using a full fledged proper HTML parser also shouldn't be a huge deal?

AliSoftware:
> As an alternative, if the need arises also for other cases, using a full fledged proper HTML parser also shouldn't be a huge deal?

Except that the PR description's body is not full HTML, but a mix of (GitHub-Flavored) Markdown with potentially embedded HTML within it. So I'm not sure throwing the PR's body text into an HTML parser would work either 🙃

spec/view_changes_need_screenshots_spec.rb
```diff
@@ -39,9 +39,18 @@ module Danger
         expect(@dangerfile.status_report[:warnings]).to be_empty
       end
 
-      it 'does nothing when a PR with view code changes has a screenshot defined with a html tag' do
+      it 'does nothing when a PR with view code changes has a screenshot defined with a html tag with different attributes before src' do
         allow(@plugin.github).to receive(:pr_body)
-          .and_return("see screenshot:\n<img width=300 src=\"https://github.com/bloom/DayOne-Apple/assets/4780/1f9e01a8-c63d-41d4-9ac8-fa9a5182ab55\"> body body")
+          .and_return("see screenshot:\n<img width=300 hint=\"First screenshots\" src=\"https://github.com/bloom/DayOne-Apple/assets/4780/1f9e01a8-c63d-41d4-9ac8-fa9a5182ab55\"> body body")
```
AliSoftware, line 44: I just realized that our RegEx would fail to detect tags that might have `>` within any of the other string attributes, like `<img alt="a > b" src="…">`…

But tbh I think that's an edge case not worth spending time on, so I'm good with it.

AliSoftware (approved):

jkmassel:
> This would require the authors of PRs to add such long markers fences around all their image tags manually every time when they write their PR description.

Ah, I had it backward, sorry – this isn't dealing with images posted by danger, it's detecting ones made by the editor. Sorry about that.

mokagio:
> Except that the PR description's body is not full HTML, but a mix of (GitHub-Flavored) Markdown with potentially embedded HTML within it. 

Is the PR description that Danger parses stored with the markup used by the author? Interesting. I don't know why, but I expected it to be the raw HTML... 🤔 I guess it could makes sense for GitHub to pass the author's version, so they don't have to bother consumers with whatever HTML structure they use to render it.

iangmaia:
> > Except that the PR description's body is not full HTML, but a mix of (GitHub-Flavored) Markdown with potentially embedded HTML within it.
> 
> Is the PR description that Danger parses stored with the markup used by the author? Interesting. I don't know why, but I expected it to be the raw HTML... 🤔 I guess it could makes sense for GitHub to pass the author's version, so they don't have to bother consumers with whatever HTML structure they use to render it.

Yes! I wasn't 100% sure and just checked, the `pr_body` in the Danger API is in Markdown with eventual HTML embedded, I'm guessing just propagating what the GitHub API returns.

AliSoftware:
I mean if we really wanted to go that route of parsing the rendered HTML from the GitHub Flavored Markdown, I guess we could use [a gem like `kramdown`](https://github.com/kramdown/parser-gfm) 🙃 

But for the time being I don't think it's worth pursuing and RegEx parsing is fine for this small use case. We can always revisit if we find out theres too many false positives and false negatives for our rule after all.

PR #30, opened 2023-11-23
Add options to check `Podfile` does not reference pods via branch

mokagio:
Our new libraries workflow expects developers to reference pods by commit and recommends not merging pods referencing a branch into `trunk`. More details at paNNhX-K4-p2.

At this time, our Danger checks prevent merging PR with pods referenced by commit while allowing merges with pods referencing branches.

This PR replicates the existing `check_podfile_does_not_have_commit_references` `check_podfile_diff_does_not_have_commit_references` but for branch references. I decided to add new methods instead of modifying the existing ones because it felt appropriate for the tool at this level to keep both options available, even though effectively we'll be using only one for the time being. As I write this, I also realize removing the existing methods would have counted as a breaking change, but the thought didn't occur at the time of developing 😅 

On top of the unit tests, I tested this live via https://github.com/wordpress-mobile/WordPress-iOS/pull/22081

See also https://github.com/Automattic/peril-settings/pull/118

### DIFF
--- a/lib/dangermattic/plugins/podfile_checks.rb
+++ b/lib/dangermattic/plugins/podfile_checks.rb
@@ -32,18 +32,11 @@ module Danger
     #
     # @return [void]
     def check_podfile_does_not_have_commit_references(podfile_lock_path: DEFAULT_PODFILE_LOCK_PATH)
-      podfile_lock_contents = File.read(podfile_lock_path)
-      podfile_lock_data = YAML.load(podfile_lock_contents)
-
-      commit_references = []
-      podfile_lock_dependencies = podfile_lock_data[PODFILE_LOCK_DEPENDENCIES_ENTRY]
-      podfile_lock_dependencies&.each do |dependency|
-        commit_references << dependency if podfile_lock_commit_reference?(podfile_line: dependency)
-      end
-
-      return if commit_references.empty?
-
-      failure("Podfile reference(s) to a commit hash:\n```#{commit_references.join("\n")}```")
+      check_podfile_does_not_match(
+        regexp: COMMIT_REFERENCE_REGEXP,
+        podfile_lock_path: podfile_lock_path,
+        match_found_message_generator: ->(matches) { "Podfile reference(s) to a commit hash:\n```#{matches.join("\n")}```" }
+      )
     end
 
     # Check for Podfile references to commit hashes in the Podfile.lock in a pull request.
@@ -51,18 +44,43 @@ module Danger
     # @return [void]
     def check_podfile_diff_does_not_have_commit_references
       warning_message = 'This PR adds a Podfile reference to a commit hash:'
-
-      git_utils.check_added_diff_lines(
-        file_selector: ->(path) { File.basename(path) == PODFILE_LOCK },
-        line_matcher: ->(line) { podfile_lock_commit_reference?(podfile_line: line) },
-        message: warning_message
-      )
+      check_podfile_diff_entries_do_not_match(regexp: COMMIT_REFERENCE_REGEXP, match_found_message: warning_message)
     end
 
     private
 
+    COMMIT_REFERENCE_REGEXP = /\(from `\S+`, commit `\S+`\)/
+
+    def check_podfile_does_not_match(
+      regexp:,
+      podfile_lock_path:,
+      match_found_message_generator: ->(matches) { "Matches found in:\n#{matches.join("\n")}" }
+    )
+      podfile_lock_contents = File.read(podfile_lock_path)
+      podfile_lock_data = YAML.load(podfile_lock_contents)
+
+      commit_references = []
+      podfile_lock_dependencies = podfile_lock_data[PODFILE_LOCK_DEPENDENCIES_ENTRY]
+      podfile_lock_dependencies&.each do |dependency|
+        commit_references << dependency if dependency.match?(regexp)
+      end
+
+      return if commit_references.empty?
+
+      failure(match_found_message_generator.call(commit_references))
+    end
+
+    def check_podfile_diff_entries_do_not_match(regexp:, match_found_message:)
+      git_utils.check_added_diff_lines(
+        # Notice the lockfile name is not configurable because we check the basename from the files in the diff and one cannot change the name of the lockfile CocoaPods generates.
+        file_selector: ->(path) { File.basename(path) == PODFILE_LOCK },
+        line_matcher: ->(line) { line.match?(regexp) },
+        message: match_found_message
+      )
+    end
+
     def podfile_lock_commit_reference?(podfile_line:)
-      podfile_line.match?(/\(from `\S+`, commit `\S+`\)/)
+      podfile_line.match?(COMMIT_REFERENCE_REGEXP)
     end
   end
 end
--- a/lib/dangermattic/plugins/podfile_checks.rb
+++ b/lib/dangermattic/plugins/podfile_checks.rb
@@ -47,9 +47,39 @@ module Danger
       check_podfile_diff_entries_do_not_match(regexp: COMMIT_REFERENCE_REGEXP, match_found_message: warning_message)
     end
 
+    # Check if the Podfile.lock contains any references to branches and raise a failure if it does.
+    #
+    # @param podfile_lock_path [String] (optional) The path to the Podfile.lock file.
+    #        Defaults to the `DEFAULT_PODFILE_LOCK_PATH` constant if not provided.
+    #
+    # @example Checking the default Podfile.lock:
+    #   check_podfile_does_not_have_branch_references
+    #
+    # @example Checking a custom Podfile.lock at a specific path:
+    #    check_podfile_does_not_have_branch_references(podfile_lock_path: '/path/to/Podfile.lock')
+    #
+    # @return [void]
+    def check_podfile_does_not_have_branch_references(podfile_lock_path: DEFAULT_PODFILE_LOCK_PATH)
+      check_podfile_does_not_match(
+        regexp: BRANCH_REFERENCE_REGEXP,
+        podfile_lock_path: podfile_lock_path,
+        match_found_message_generator: ->(matches) { "Podfile reference(s) to a branch:\n```#{matches.join("\n")}```" }
+      )
+    end
+
+    # Check for Podfile references to branches in the Podfile.lock in a pull request.
+    #
+    # @return [void]
+    def check_podfile_diff_does_not_have_branch_references
+      warning_message = 'This PR adds a Podfile reference to a branch:'
+      check_podfile_diff_entries_do_not_match(regexp: BRANCH_REFERENCE_REGEXP, match_found_message: warning_message)
+    end
+
     private
 
     COMMIT_REFERENCE_REGEXP = /\(from `\S+`, commit `\S+`\)/
+
+    BRANCH_REFERENCE_REGEXP = /\(from `\S+`, branch `\S+`\)/
 
     def check_podfile_does_not_match(
       regexp:,
--- a/lib/dangermattic/plugins/podfile_checks.rb
+++ b/lib/dangermattic/plugins/podfile_checks.rb
@@ -108,9 +108,5 @@ module Danger
         message: match_found_message
       )
     end
-
-    def podfile_lock_commit_reference?(podfile_line:)
-      podfile_line.match?(COMMIT_REFERENCE_REGEXP)
-    end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/dangermattic/plugins/podfile_checks.rb
```diff
@@ -32,37 +32,85 @@ class PodfileChecks < Plugin
...
+      )
+    end
+
+    # Check for Podfile references to commit hashes in the Podfile.lock in a pull request.
+    #
+    # @return [void]
+    def check_podfile_diff_does_not_have_commit_references
+      warning_message = 'This PR adds a Podfile reference to a commit hash:'
+      check_podfile_diff_entries_do_not_match(regexp: COMMIT_REFERENCE_REGEXP, match_found_message: warning_message)
+    end
```
mokagio, line 48: I basically just abstracted the previous implementation to inject the message and the RegExp then used different pairs for commit and branch checks.

lib/dangermattic/plugins/podfile_checks.rb
```diff
@@ -32,37 +32,85 @@ class PodfileChecks < Plugin
...
-        line_matcher: ->(line) { podfile_lock_commit_reference?(podfile_line: line) },
-        message: warning_message
+        line_matcher: ->(line) { line.match?(regexp) },
+        message: match_found_message
       )
     end
 
-    private
-
     def podfile_lock_commit_reference?(podfile_line:)
```
iangmaia: I believe `podfile_lock_commit_reference?` can be completely removed now, as you use the common regex matching?

lib/dangermattic/plugins/podfile_checks.rb
iangmaia: Great idea and resulting implementation 👍

lib/dangermattic/plugins/podfile_checks.rb
```diff
@@ -32,37 +32,85 @@ class PodfileChecks < Plugin
...
+    #        Defaults to the `DEFAULT_PODFILE_LOCK_PATH` constant if not provided.
+    #
+    # @example Checking the default Podfile.lock:
+    #   check_podfile_does_not_have_branch_references
+    #
+    # @example Checking a custom Podfile.lock at a specific path:
+    #    check_podfile_does_not_have_branch_references(podfile_lock_path: '/path/to/Podfile.lock')
+    #
+    # @return [void]
+    def check_podfile_does_not_have_branch_references(podfile_lock_path: DEFAULT_PODFILE_LOCK_PATH)
```
iangmaia, line 62: This was already present in the previous implementation and, to be honest, I cannot recall what was the precise original idea, but the full `Podfile` check reports an error and the `diff` check reports a warning. ~Probably just a verbatim of Peril's current behaviour.~ [**Update**: when reviewing Automattic/peril-settings/pull/118 , I noticed Peril only does the full file check, always failing. I added the diff check for consistency (also, it is a kind of pattern for Danger Plugins to offer both).]

I have a task for myself to review all plugins regarding error/warning reporting, perhaps adding the option for either option in more of them, or making some of them more strictly reporting errors, but perhaps, while at it, we can already give this case some thought.

What's your take for this plugin (and in general)?

If we want to give users the option to choose between error / warning with a default value, it should be rather simple (`git_utils.check_added_diff_lines()` already provides a `fail_on_error: false` parameter).

spec/podfile_checks_spec.rb
```diff
@@ -298,6 +299,299 @@ module Danger
...
+
+            COCOAPODS: 1.12.1
+          CONTENT
+
+          allow(File).to receive(:read).with('./Podfile.lock').and_return(podfile_lock_content)
+
+          @plugin.check_podfile_does_not_have_branch_references
+
+          expected_message = "Podfile reference(s) to a branch:\n```Kingfisher (from `https://github.com/onevcat/Kingfisher.git`, branch `main`)```"
+          expect(@dangerfile.status_report[:errors]).to eq [expected_message]
```
iangmaia:
```suggestion
          expect(@dangerfile).to report_errors([expected_message])
```

spec/podfile_checks_spec.rb
```diff
@@ -298,6 +299,299 @@ module Danger
...
+
+          @plugin.check_podfile_does_not_have_branch_references
+
+          expected_message = <<~MESSAGE.chomp
+            Podfile reference(s) to a branch:
+            ```Kingfisher (from `https://github.com/onevcat/Kingfisher.git`, branch `main`)
+            SwiftGen (from `https://github.com/SwiftGen/SwiftGen.git`, branch `test/a-test-branch`)
+            SwiftLint (from `https://github.com/realm/SwiftLint.git`, branch `feature-branch`)```
+          MESSAGE
+          expect(@dangerfile.status_report[:errors]).to eq [expected_message]
```
iangmaia:
```suggestion
          expect(@dangerfile).to report_errors([expected_message])
```

spec/podfile_checks_spec.rb
```diff
@@ -298,6 +299,299 @@ module Danger
...
+            COCOAPODS: 1.12.1
+          PODFILE
+
+          diff = GitDiffStruct.new('modified', podfile_path, podfile_lock_diff)
+
+          allow(@plugin.git).to receive(:diff_for_file).with(podfile_path).and_return(diff)
+
+          @plugin.check_podfile_diff_does_not_have_branch_references
+
+          expect(@dangerfile.status_report[:warnings]).to be_empty
```
iangmaia:
```suggestion
          expect(@dangerfile).to not_report
```

spec/podfile_checks_spec.rb
```diff
@@ -298,6 +299,299 @@ module Danger
...
+            PODFILE CHECKSUM: 58bea7d9e64aa93550617dd989b9e5ceed3df3f4
+
+            COCOAPODS: 1.12.1
+          CONTENT
+
+          allow(File).to receive(:read).with('./Podfile.lock').and_return(podfile_lock_content)
+
+          @plugin.check_podfile_does_not_have_branch_references
+
+          expect(@dangerfile.status_report[:errors]).to be_empty
```
iangmaia:
```suggestion
          expect(@dangerfile).to not_report
```

spec/podfile_checks_spec.rb
```diff
@@ -298,6 +299,299 @@ module Danger
...
+
+          expected_warning2 = <<~WARNING
+            This PR adds a Podfile reference to a branch:
+            File `#{podfile_path}`:
+            ```diff
+            +  - SwiftLint (from `https://github.com/realm/SwiftLint.git`, branch `develop`)
+            ```
+          WARNING
+
+          expect(@dangerfile.status_report[:warnings]).to contain_exactly(expected_warning, expected_warning2)
```
iangmaia:
```suggestion
          expect(@dangerfile).to report_warnings([expected_warning, expected_warning2])
```

spec/podfile_checks_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 require_relative 'spec_helper'
 
+# rubocop:disable Metrics/ModuleLength
```
iangmaia, line 5: Perhaps worth disabling for specs in the `.rubocop.yml`? 🤔

iangmaia:
@mokagio pinging so that we can land this and apply it according to paaHJt-5zM-p2 🙂

lib/dangermattic/plugins/podfile_checks.rb
mokagio: Good question. I'll admit I didn't think through the design too much while working on this PR as I was motivated by getting it done ASAP—which is ironic in hindsight given I let it go stale 😳 

I think we have two questions here, right?

- Should the `Podfile` (and, generalizing, any lockfile) check inspect the entire file or the PR diff only?
- Should the check report a warning or an error?

Plus, the design question of whether to add a `fail_on_error:` configuration for the caller.

I'll start with the easiest one for me to answer. 

#### `fail_on_error`?

I'm 100% on board with adding a `fail_on_error:` param with a reasonable default to this and every other check. I think it aligns with the guiding principle of building _flexible tools with opinionated golden paths_.

In writing this I briefly considered suggesting even more flexibility, e.g. (`on_violation: :fail # :warn, :comment`). That would be neat, but I think that most of our checks work to provide the user feedback in terms of warnings or straight out failures, with the comment option used by only a few checks. Example from Peril:

image

Under that assumption, it seems that passing and handling a `Bool` would be much simpler that establishing and maintaining a consistent `:fail, :warn, :comment` pattern.

The only caveat I have is that I'd like to use the word "violation" instead of "error". That is `fail_on_violation: true` over `fail_on_error: true`. I feel like "error" is too harsh for the kind of check failures that Danger can pick up. _I also acknowledge this is overthinking it and it's easier to update. However, given we have many clients I think it's worth spending extra time to settle on something less likely to need changes in the future._

What do you think?

### Check the full lockfile or the diff only?

The case for checking the diff only is that Danger is a tool for "unit testing pull requests" so checking the diff seems like the most appropriate choice, given that PRs modify files in diffs.

Based on that rationale, I'd say to keep only the diff check. Even though intuitively I would expect Danger to check the whole file in this instance.

But if we generalize the diff vs whole file behavior, then it seems clear to me that diff is the way to go. Let's consider a linter like RuboCop. It would be rather noisy to get warnings about pre-existing violations that a PR did not introduce, despite the fact that those violations ought to be addressed ASAP.

So, for the sake of consistency, I think diff-only is the approach we should take here, too.

### Warn or fail when user fetches pods by commit?

The operating procedure we agree upon, paNNhX-K4-p2, allow references to commits to be used during development, so I think a warning to raise awareness is enough here.

...which just made me realize that the check for branch references which I considered useless and kept just for reference actually becomes more useful that the check for commit. Because, while using commits is okay except before making a release build (which we could implement by changing the `fail_on_violation` value depending on the base branch) referencing branches should never be allowed, we can therefore add that check, too to our `Dangerfile` with `fail_on_violation: true`.

iangmaia: > The only caveat I have is that I'd like to use the word "violation" instead of "error". That is `fail_on_violation: true` over `fail_on_error: true`. I feel like "error" is too harsh for the kind of check failures that Danger can pick up. _I also acknowledge this is overthinking it and it's easier to update. However, given we have many clients I think it's worth spending extra time to settle on something less likely to need changes in the future._
> 
> What do you think?

Agreed! Plus, when `fail_on_error` is `false`, it's not really an "error" if we're going to report a warning 😄 
A parameter with the options `:fail`, `:warn`, `:comment` also isn't a bad idea -- it's something [we implemented](https://github.com/Automattic/dangermattic/blob/4df9a409f6a5d78a63b55d6e5fb8bdb27a5efd8a/lib/dangermattic/plugins/milestone_checker.rb#L59) in another plugin a slightly different purpose.

What about calling it `severity:` or `report:` with the  options `:fail`, `:warn`, `:comment`?
I'll work in the coming days on a set of improvements for Dangermattic, and one of them will be on error reporting, so I can pick this up.

> ### Check the full lockfile or the diff only?
. [...]
> But if we generalize the diff vs whole file behavior, then it seems clear to me that diff is the way to go. Let's consider a linter like RuboCop. It would be rather noisy to get warnings about pre-existing violations that a PR did not introduce, despite the fact that those violations ought to be addressed ASAP.

That's also the thing: Danger RuboCop and Danger SwiftLint both offer the option to lint everything or lint only the diff, and perhaps this philosophy is worth incorporating as well, at least in some cases.

> ### Warn or fail when user fetches pods by commit?
> The operating procedure we agree upon, [paNNhX-K4-p2](https://href.li/?https://wp.me/paNNhX-K4), allow references to commits to be used during development, so I think a warning to raise awareness is enough here.

👍 

> ...which just made me realize that the check for branch references which I considered useless and kept just for reference actually becomes more useful that the check for commit. Because, while using commits is okay except before making a release build (which we could implement by changing the `fail_on_violation` value depending on the base branch) referencing branches should never be allowed, we can therefore add that check, too to our `Dangerfile` with `fail_on_violation: true`.

Sounds good 👍

mokagio: @iangmaia thank you for promptly following up—_not like I do_. 

Based on the discussion above, is there anything left to address in this PR, or can we merge it?

spec/podfile_checks_spec.rb
mokagio: https://github.com/Automattic/dangermattic/pull/35

iangmaia (approved):

lib/dangermattic/plugins/podfile_checks.rb
iangmaia: Let's merge it! :shipit:

PR #31, opened 2023-12-23
Reusable Workflow for checking GitHub Issues labels

iangmaia:
This PR implements a reusable workflow for checking the labels configured on GitHub Issues.
It contains a generic label verification logic that can be parameterised, so that [these Peril checks](https://github.com/Automattic/peril-settings/blob/master/org/issue/label-woo.ts) can be reimplemented on GitHub Actions. 

## How to test
The best way to test this is to use the workflow in another repo _[please remember that GitHub Issue events are only triggered once a workflow is on the main branch]_.

For example, [this is a fork of the WooCommerce App on Android with the shared workflow being used](https://github.com/iangmaia/woocommerce-android/blob/trunk/.github/workflows/labels-on-issues.yml). Feel free to test issues over there.

### DIFF
--- a/.github/workflows/reusable-check-labels-on-issues.yml
+++ b/.github/workflows/reusable-check-labels-on-issues.yml
@@ -1,0 +1,91 @@
+on:
+  workflow_call:
+    inputs:
+      label-format-list:
+        description: 'The Regex formats expected for the labels; must be a JSON list'
+        default: '[
+          ".*"
+        ]'
+        type: string
+        required: false
+      label-error-message:
+        description: 'Error message to be posted when the labels set don''t match the required list of formats.'
+        default: 'At least one label is required.'
+        type: string
+        required: false
+      label-success-message:
+        description: 'Message to be posted when the labels set fulfill the entire list of expected formats.'
+        default: '✅ Yay, issue looks great!'
+        type: string
+        required: false
+    secrets:
+      github-token:
+        required: true
+
+concurrency:
+  group: danger-${{ github.event.issue.number }}
+  cancel-in-progress: true
+
+jobs:
+  check-issue-labels:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: "🏷️ Check Issue Labels"
+        env:
+          GITHUB_TOKEN: ${{ secrets.github-token }}
+          GH_REPO: ${{ github.repository }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          ISSUE_LABELS: ${{ toJSON(github.event.issue.labels.*.name) }}
+          LABEL_REGEX_LIST: ${{ inputs.label-format-list }}
+          ISSUE_SUCCESS_COMMENT: >
+            ${{ inputs.label-success-message }}
+            <!-- generated_by_dangermattic -->
+          ISSUE_ERROR_COMMENT: >
+            ${{ inputs.label-error-message }}
+            <!-- generated_by_dangermattic -->
+        run: |
+          #!/bin/bash
+
+          readarray -t labels < <(echo "$ISSUE_LABELS" | jq -r '.[]')
+          readarray -t label_regex_list < <(echo "$LABEL_REGEX_LIST" | jq -r '.[]')
+
+          all_patterns_matched=true
+
+          for regex in "${label_regex_list[@]}"; do
+            pattern_matched=false
+
+            for label in "${labels[@]}"; do
+              if [[ "$label" =~ $regex ]]; then
+                pattern_matched=true
+                echo "👍 Match found for regex '$regex': $label"
+                break
+              fi
+            done
+
+            if [ "$pattern_matched" = false ]; then
+              all_patterns_matched=false
+              echo "⚠️ No match found for regex '$regex'"
+            fi
+          done
+
+          ISSUE_COMMENT=''
+          if [ "$all_patterns_matched" = true ]; then
+            echo "✅ All regex patterns have at least one match."
+            ISSUE_COMMENT="$ISSUE_SUCCESS_COMMENT"
+          else
+            echo "❌ Not all regex patterns have at least one match."
+            ISSUE_COMMENT="$ISSUE_ERROR_COMMENT"
+          fi
+
+          set +e
+          echo "✍️ Attempting to edit existing comment on issue $ISSUE_NUMBER, if it exists:"
+          gh issue comment $ISSUE_NUMBER --body "$ISSUE_COMMENT" --edit-last
+          comment_update_status=$?
+          set -e
+
+          if [ $comment_update_status -ne 0 ]; then
+            echo "✍️ Adding new comment on issue $ISSUE_NUMBER:"
+            gh issue comment $ISSUE_NUMBER --body "$ISSUE_COMMENT"
+          fi


AliSoftware:
It's a bit of a shame to have a Danger plugin implemented in dangermattic for exactly the same logic already (but for PRs), but for issues we have to go with a completely separate implementation (writing the logic in bash within the GHA).

This is not gonna be ideal for long term maintenance if we have two places where this labels-checking issue is implemented 😞  Any idea if there's a way around that?

 - Like running `danger` on issues (even if it means passing a special flag for it not to try resolving things like "base and head branch" and such so it doesn't crash)
 - Or maybe extracting the common implementation about labels check in a separate `check_labels.rb` Ruby file that could be called stand-alone, and have this GHA on issues call `ruby check_labels.rb`, while the Danger plugin working on PRs would be just a Danger Action wrapper around `check_labels.rb`?

On the other end that might be overkill given the bash implementation is not that long… though arguably still not as readable as Ruby code, potentially easier to get it wrong given bash tricks, and not unit-tested either, so maybe still valuable.

crazytonyli:
Here is an idea on top of my head regarding reusing the ruby danger code:
1. Implement the checks in danger ruby. Let's say it's called `issue_labels_check`.
2. In the GitHub action shell script, create a temporary file `/tmp/issue-xxx.dangerfile` with content like this:
```
issue_labels_chekc.do_your_thing
issue_labels_check.report
```
3. Finally, in the same shell script, call `bundle exec danger pr --danger-file /tmp/issue-xxx.dangerfile`.

Would it work?

AliSoftware:
> Would it work?

@crazytonyli I believe that the core of the issue is not really that we need a separate Dangerfile for issues (we could do that even without creating a temp file, like having `Dangerfile-Issues` commited in the repo next to `Dangerfile`), but the fact that `danger` doesn't seem to work well with issues in the first place.

Indeed, the `danger` tool was implemented with PRs in mind in the first place, and if you try to run it on the URL of an issue, it will crash failing to find the `base` and `head` branch associated with the issue or other PR-related metadata (given it will wrongly expect it to be a PR).

That being said, maybe there's an option or trick I don't know about in `danger` to allow it to work on issues and handle them properly? Or if not, maybe it's worth checking if someone hasn't made a PR in danger repo to add support to that, that we could contribute to and help make it land?

@iangmaia did you try tricking danger by calling it on an issue but with `--base=trunk --head=trunk` (or maybe trying with two different dummy branches, in case it doesn't like having base==head), to see if it might help avoid the crash of it trying to fetch base/head info? Or is it crashing elsewhere anyway?

iangmaia:
Thanks for the input @crazytonyli and @AliSoftware! 🙇 

> but the fact that `danger` doesn't seem to work well with issues in the first place

🎯 
I realised I should have made it explicit in the original RFC paaHJt-5TT-p2 / in this PR description, but the main problem that makes us going this route is Danger Ruby not really supporting Issues.


> Indeed, the `danger` tool was implemented with PRs in mind in the first place, and if you try to run it on the URL of an issue, it will crash failing to find the `base` and `head` branch associated with the issue or other PR-related metadata (given it will wrongly expect it to be a PR).
> 
> That being said, maybe there's an option or trick I don't know about in `danger` to allow it to work on issues and handle them properly? Or if not, maybe it's worth checking if someone hasn't made a PR in danger repo to add support to that, that we could contribute to and help make it land?
> 
> @iangmaia did you try tricking danger by calling it on an issue but with `--base=trunk --head=trunk` (or maybe trying with two different dummy branches, in case it doesn't like having base==head), to see if it might help avoid the crash of it trying to fetch base/head info? Or is it crashing elsewhere anyway?

I think there's also a problem on the entry point for Danger, as usually there isn't CI running for issues. At this point, Danger will try to get some data from the CI environment and fetch more info.
I've tried to [fork Danger](https://github.com/danger/danger/compare/master...iangmaia:danger:danger-on-gh-issues) to see how big of an effort would it be to have it working but eventually abandoned this idea, as it seemed it would require a bigger refactoring and our need here is basically issues label validation.
I think that's what's key here: our use case is rather simple to justify something more complete / flexible.


> Or maybe extracting the common implementation about labels check in a separate check_labels.rb Ruby file that could be called stand-alone, and have this GHA on issues call ruby check_labels.rb, while the Danger plugin working on PRs would be just a Danger Action wrapper around check_labels.rb?

This is an approach I was more seriously considering. There are a couple of problems (though not very severe) such as 1) the distribution of this common code and 2) the Ruby + dependencies (at least Octokit) setup required by this script, as well as the project checkout, also adding up slightly the GHA execution time for Issues.
For the time being I just went for the bash script which isn't too complex, has no dependencies and was simple enough to reuse using this shared workflow.

.github/workflows/reusable-check-labels-on-issues.yml
```diff
@@ -0,0 +1,91 @@
...
+    inputs:
+      label-format-list:
+        description: 'The Regex formats expected for the labels; must be a JSON list'
+        default: '[
+          ".*"
+        ]'
+        type: string
+        required: false
+      label-error-message:
+        description: 'Error message to be posted when the labels set don''t match the required list of formats.'
```
mokagio, line 12: TIL you can do that trick with the two `''`

.github/workflows/reusable-check-labels-on-issues.yml
```diff
@@ -0,0 +1,91 @@
...
+        default: 'At least one label is required.'
+        type: string
+        required: false
+      label-success-message:
+        description: 'Message to be posted when the labels set fulfill the entire list of expected formats.'
+        default: '✅ Yay, issue looks great!'
```
mokagio, line 18: Do we _need_ a message if everything looks fine?

.github/workflows/reusable-check-labels-on-issues.yml
```diff
@@ -0,0 +1,91 @@
...
+          set +e
+          echo "✍️ Attempting to edit existing comment on issue $ISSUE_NUMBER, if it exists:"
+          gh issue comment $ISSUE_NUMBER --body "$ISSUE_COMMENT" --edit-last
+          comment_update_status=$?
+          set -e
+
+          if [ $comment_update_status -ne 0 ]; then
+            echo "✍️ Adding new comment on issue $ISSUE_NUMBER:"
+            gh issue comment $ISSUE_NUMBER --body "$ISSUE_COMMENT"
+          fi
```
mokagio, line 91: Have you considered exiting with `comment_update_status`, so that if something fails at the `gh` level there are more chances we'll notice?

mokagio (commented):
This is a neat solution to the problem of how to run the kind of checks we want for issue without Danger.

Have you got any ideas on how to extract the script part so it's a bit easier to work on (and maybe unit test 🤞) than as a multiline string in a YAML file?

mokagio (approved):
I tried this by creating a new issue in the test repo, https://github.com/iangmaia/woocommerce-android/issues/5, and playing with the labels.

- https://github.com/iangmaia/woocommerce-android/actions/runs/7551438861
- https://github.com/iangmaia/woocommerce-android/actions/runs/7551433869
- https://github.com/iangmaia/woocommerce-android/actions/runs/7551455967
- https://github.com/iangmaia/woocommerce-android/actions/runs/7551460795

I think from the comments we all agree that this approach is far from what we'd like to have. 

But I think that **if the only problem we are trying to solve is something as straightforward and unlikely to change much over time as commenting on issues based on their labels and milestones**, then this does the job alright and we can "set and forget".

.github/workflows/reusable-check-labels-on-issues.yml
iangmaia: Good point 👍

.github/workflows/reusable-check-labels-on-issues.yml
iangmaia: True, we don't need a `default` here.

.github/workflows/reusable-check-labels-on-issues.yml
iangmaia: Oh, actually the first error is part of a happy path: it's in the case when there's no comment yet and we tried to edit one.

.github/workflows/reusable-check-labels-on-issues.yml
iangmaia: Again going back to this I realise why 😓 

The API via the `gh` CLI is rather limited. It doesn't allow for fetching comments or deleting comments, only adding a comment or editing the latest one. To make the code simpler, avoiding curl and more calls, I use just what's available and that seemed ok for this use case.

iangmaia:
@mokagio:
> Have you got any ideas on how to extract the script part so it's a bit easier to work on (and maybe unit test 🤞) than as a multiline string in a YAML file?

Bash in YAML workflow files seems like a common thing in the GHA world 😅 
But the "problem" here is the fact that this is a reusable workflow: things happening in this workflow are actually happening in the caller workflow. So if we move the script to an external file or a Ruby script, we'd need to add it to all repos using this reusable workflow (in fact, this is how the [reusable GHA Danger workflow](https://github.com/Automattic/dangermattic/blob/trunk/.github/workflows/reusable-run-danger.yml) works).

iangmaia:
> I think from the comments we all agree that this approach is far from what we'd like to have.
> 
> But I think that if the only problem we are trying to solve is something as straightforward and unlikely to change much over time as commenting on issues based on their labels and milestones, then this does the job alright and we can "set and forget".

Exactly my thinking 👍

mokagio:
@iangmaia 

> But the "problem" here is the fact that this is a reusable workflow: things happening in this workflow are actually happening in the caller workflow. So if we move the script to an external file or a Ruby script, we'd need to add it to all repos using this reusable workflow

I was thinking of having something like source files and test files, then a deploy script that code generates the YML. The big, very big, drawback of that approach is that one would have to remember to run the code generation and that the YML would have to be tracked in Git to be used as a workflow. That makes is an annoying solution in my books. There could be ways to make it work with Git hooks and the likes, but I'd say that if the amount of logic we have is this minimal and seldom changing, we're good like this 👍

AliSoftware:
> I was thinking of having something like source files and test files, then a deploy script that code generates the YML. The big, very big, drawback of that approach is that one would have to remember to run the code generation and that the YML would have to be tracked in Git to be used as a workflow. That makes is an annoying solution in my books. There could be ways to make it work with Git hooks and the likes, but I'd say that if the amount of logic we have is this minimal and seldom changing, we're good like this 👍

@mokagio what about the opposite? Have the test code first parse the YAML workflow, extract `jobs.check-issue-labels.steps[0].run` from that YAML, and run the tests (like `shellcheck` as well as testing code) on the extracted shell script?

That way, no need to ensure the YAML is in sync with the source code or to remember to re-generate the YML if we update the script.

mokagio:
> what about the opposite?

@AliSoftware that's a rad idea!

PR #35, opened 2024-01-17
Exclude specs from the `Metrics/ModuleLength` linter

mokagio:
Specs can be verbose and end up being quite long, but that's not a problem in the context of unit tests. 

That's not to say we should not strive to keep our tests readable and tidy, but their maintainability depends less on line length (a rough metric, anyways) than the production code does.


iangmaia (approved):
👍

Issue #29, opened 2023-11-15
Unclear warning about merging PR before milestone date


crazytonyli:
> This PR is assigned to a milestone which is closing in less than 4 days Please, make sure to get it merged by then or assign it to a later expiring milestone

This warning was added to a PR whose base branch is code-frozen release branch, which is obviously already past the milestone date. The "make sure to get it merged by then" part seems a bit unclear, if "then" means the milestone code freeze date.

See https://github.com/wordpress-mobile/WordPress-iOS/pull/22041#issuecomment-1811527388.

iangmaia:
Thanks for reporting it @crazytonyli, I remember also seeing this scenario somewhere else 👍

iangmaia:
Fixed in #40

PR #40, opened 2024-01-30
Plugins Error Policy Review

iangmaia:
Previously, Dangermattic had a pattern on the public API to allow for some customization on error levels reporting: the parameter `fail_on_error [Boolean]`, though this wasn't strictly followed on all plugins.

This PR changes this pattern to a `report_type [Symbol]` parameter instead. I've also added this parameter on all APIs that I thought made sense. This is based on a few discussions I've had on this topic in other PRs (e.g. [last one](https://github.com/Automattic/dangermattic/pull/30#discussion_r1452888075) with @mokagio).
 
The reporting itself is implemented in the `Reporter` class, created as a plugin for easy access from other plugins.

### DIFF
--- a/lib/dangermattic/plugins/common/git_utils.rb
+++ b/lib/dangermattic/plugins/common/git_utils.rb
@@ -45,7 +45,7 @@ module Danger
     #
     # @param message [String] The warning or failure message to display when the pattern is found in a line.
     #
-    # @param report_type [Boolean] (optional) Type of report (:error, :warning, :message) whenever a line matches the criteria. Default is :warning.
+    # @param report_type [Symbol] (optional) Type of report (:error, :warning, :message) whenever a line matches the criteria. Default is :warning.
     #
     # @example Checking for added lines containing 'FIXME' and failing the build:
     #   check_added_diff_lines(file_selector: ->(path) { File.extname(path) == ('.swift') }, line_matcher: ->(line) { line.include?("FIXME") }, message: "A FIXME was added, failing build.", report_type: :error)
--- a/lib/dangermattic/plugins/common/git_utils.rb
+++ b/lib/dangermattic/plugins/common/git_utils.rb
@@ -45,13 +45,13 @@ module Danger
     #
     # @param message [String] The warning or failure message to display when the pattern is found in a line.
     #
-    # @param fail_on_error [Boolean] (optional) When set to true, whenever a line matches the criteria, the method will emit an error, when false, emit a warning. Default is false.
+    # @param report_type [Boolean] (optional) When set to true, whenever a line matches the criteria, the method will emit an error, when false, emit a warning. Default is false.
     #
     # @example Checking for added lines containing 'FIXME' and failing the build:
-    #   check_added_diff_lines(file_selector: ->(path) { File.extname(path) == ('.swift') }, line_matcher: ->(line) { line.include?("FIXME") }, message: "A FIXME was added, failing build.", fail_on_error: true)
+    #   check_added_diff_lines(file_selector: ->(path) { File.extname(path) == ('.swift') }, line_matcher: ->(line) { line.include?("FIXME") }, message: "A FIXME was added, failing build.", report_type: :error)
     #
     # @return [void]
-    def check_added_diff_lines(file_selector:, line_matcher:, message:, fail_on_error: false)
+    def check_added_diff_lines(file_selector:, line_matcher:, message:, report_type: :warning)
       modified_files = added_and_modified_files.select(&file_selector)
 
       matches = matching_lines_in_diff_files(
@@ -70,11 +70,7 @@ module Danger
             ```
           MESSAGE
 
-          if fail_on_error
-            failure(final_message)
-          else
-            warn(final_message)
-          end
+          reporter.report(message: final_message, type: report_type)
         end
       end
     end
--- a/lib/dangermattic/plugins/common/git_utils.rb
+++ b/lib/dangermattic/plugins/common/git_utils.rb
@@ -45,7 +45,7 @@ module Danger
     #
     # @param message [String] The warning or failure message to display when the pattern is found in a line.
     #
-    # @param report_type [Boolean] (optional) When set to true, whenever a line matches the criteria, the method will emit an error, when false, emit a warning. Default is false.
+    # @param report_type [Boolean] (optional) Type of report (:error, :warning, :message) whenever a line matches the criteria. Default is :warning.
     #
     # @example Checking for added lines containing 'FIXME' and failing the build:
     #   check_added_diff_lines(file_selector: ->(path) { File.extname(path) == ('.swift') }, line_matcher: ->(line) { line.include?("FIXME") }, message: "A FIXME was added, failing build.", report_type: :error)
--- a/lib/dangermattic/plugins/common/reporter.rb
+++ b/lib/dangermattic/plugins/common/reporter.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module Danger
+  # Handles reporting messages in the context of Danger, allowing the generation of
+  # warnings, errors, and simple messages in a Danger run.
+  #
+  # @example Reporting a Warning
+  #   reporter.report(message: "This is a warning message", type: :warning)
+  #
+  # @example Reporting an Error
+  #   reporter.report(message: "This is an error message", type: :error)
+  #
+  # @see Automattic/dangermattic
+  # @tags tool, util, danger
+  #
+  class Reporter < Plugin
+    # Report a message to be posted by Danger as an error (failing the build), a warning or a simple message.
+    #
+    # @param message [String] The message to be reported to Danger.
+    # @param type [Symbol] The type of report. Possible values:
+    #                        - :warning (default): Reports a warning.
+    #                        - :error: Reports an error.
+    #                        - :message: Reports a simple message.
+    #                        - :none or nil: Takes no action.
+    #
+    # @return [void]
+    def report(message:, type: :warning)
+      return if message.nil? || message.empty? || type.nil?
+
+      case type
+      when :error
+        failure(message)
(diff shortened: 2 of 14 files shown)


lib/dangermattic/plugins/milestone_checker.rb
```diff
@@ -77,13 +68,14 @@ def check_milestone_due_date(days_before_due: DEFAULT_DAYS_THRESHOLD, if_no_mile
 
       message_text = "This PR is assigned to the milestone [#{milestone_title}](#{milestone_url}). "
       message_text += if time_before_due_date.positive?
-                        "This milestone is due in less than #{days_before_due} days.\n"
+                        "This milestone is due in less than #{days_before_due} days.\n" \
```
iangmaia, line 68: While at it, I've improved the message on the Milestones plugin, which was reported as an issue (https://github.com/Automattic/dangermattic/issues/29) by @crazytonyli.

spec/milestone_checker_spec.rb
```diff
@@ -186,7 +186,7 @@ module Danger
         it "reports a warning when asked to do so when a PR doesn't have a milestone set" do
           allow(@plugin.github).to receive(:pr_json).and_return({})
 
-          @plugin.check_milestone_due_date(if_no_milestone: :warn)
+          @plugin.check_milestone_due_date(report_if_no_milestone: :warning)
```
mokagio: `report_if_no_milestone` sounds like a yes/no question to me.

With Ruby being a weakly typed language, my first instinct would be to pass `true` and I would learn it's incorrect only later on (_this of course assumes my text editor / IDE is not smart enough to check that for me and that I jumped into coding without checking the API docs or source..._)

What do you think of taking a page from the [Swift API Guidelines](https://swift.org/documentation/api-design-guidelines/#weak-type-information) recommendation to precede each weakly typed parameter, such as `Any`, `Int`, or `String, with a noun describing its role:

screen shot 2019-02-06 at 1 43 58 pm

```diff
-          @plugin.check_milestone_due_date(report_if_no_milestone: :warning)
+         @plugin.check_milestone_due_date(report_type_if_no_milestone: :warning)
```

lib/dangermattic/plugins/common/reporter.rb
```diff
@@ -0,0 +1,40 @@
...
+  # @tags tool, util, danger
+  #
+  class Reporter < Plugin
+    # Report a message to be posted by Danger as an error (failing the build), a warning or a simple message.
+    #
+    # @param message [String] The message to be reported to Danger.
+    # @param type [Symbol] The type of report. Possible values:
+    #                        - :warning (default): Reports a warning.
+    #                        - :error: Reports an error.
+    #                        - :message: Reports a simple message.
```
mokagio, line 23: Nitpick. I felt unbalanced having `message` as the argument label and `:message` as a report type.

I thought about it a bit but couldn't find anything better than `text`. AI suggested `content`, which is not that much better.

I looked at the [Danger source](https://github.com/danger/danger/blob/bd89c4d6f3c7650d766eb804000a10f0b2fb98ab/lib/danger/request_sources/github/github_review_unsupported.rb#L13-L19) and they seem to be fine with the same clash. So maybe I'm overthinking it, as it's often the case.

lib/dangermattic/plugins/common/reporter.rb
```diff
@@ -0,0 +1,40 @@
...
+    #                        - :none or nil: Takes no action.
+    #
+    # @return [void]
+    def report(message:, type: :warning)
+      return if message.nil? || message.empty? || type.nil?
+
```
mokagio: Another nipick. What do you think of trusting danger to know how to best handle `nil` and empty messages?

If so, we could simplify and remove this whole check, because if `type` is `nil` the `case` will simply not run into any of the `when` branches below.

mokagio (commented):
I got interrupted. Here's my partial thoughts. Looking good so far, anyway

spec/milestone_checker_spec.rb
iangmaia: Good point! Fixed on 5db5ee68fa5a4649694535eee6947f5cb1d45fbd.

lib/dangermattic/plugins/common/reporter.rb
iangmaia: I see what you mean -- I had this problem myself here on another level, with the local var `message` and Danger's function `message()`, which I had then to call using `danger.message()`.

The problem is, I think `message` is probably what's best describing what these things are, given Danger's nomenclature 🤷

lib/dangermattic/plugins/common/reporter.rb
iangmaia: [Checked](https://github.com/danger/danger/blob/bd89c4d6f3c7650d766eb804000a10f0b2fb98ab/lib/danger/danger_core/plugins/dangerfile_messaging_plugin.rb#L126) just in case 🙂  But yeah, I think it can be worth simplifying. Changed on dfd264dd26db273b4a6f825f7d475f23b6063b4b

lib/dangermattic/plugins/common/reporter.rb
```diff
@@ -25,8 +25,6 @@ class Reporter < Plugin
     #
     # @return [void]
     def report(message:, type: :warning)
-      return if message.nil? || message.empty? || type.nil?
-
```
mokagio: Thank you for following up on this.

If the behavior changes on Danger's side somehow, or if making sure the behavior on our hand is robust becomes more important, we can always add some tests around what Danger itself does and run them as part of our unit tests to make sure it's consistent with our assumptions. 

But that sounds like overkill at this point in time.

lib/dangermattic/plugins/common/reporter.rb
mokagio: > I think message is probably what's best describing what these things are

Yeah, same. 🤷‍♂️

lib/dangermattic/plugins/common/git_utils.rb
```diff
@@ -45,13 +45,13 @@ class GitUtils < Plugin
     #
     # @param message [String] The warning or failure message to display when the pattern is found in a line.
     #
-    # @param fail_on_error [Boolean] (optional) When set to true, whenever a line matches the criteria, the method will emit an error, when false, emit a warning. Default is false.
+    # @param report_type [Boolean] (optional) Type of report (:error, :warning, :message) whenever a line matches the criteria. Default is :warning.
```
mokagio:
```suggestion
    # @param report_type [Symbol] (optional) Type of report (:error, :warning, :message) whenever a line matches the criteria. Default is :warning.
```

lib/dangermattic/plugins/common/reporter.rb
```diff
@@ -0,0 +1,38 @@
...
+  #
+  class Reporter < Plugin
+    # Report a message to be posted by Danger as an error (failing the build), a warning or a simple message.
+    #
+    # @param message [String] The message to be reported to Danger.
+    # @param type [Symbol] The type of report. Possible values:
+    #                        - :warning (default): Reports a warning.
+    #                        - :error: Reports an error.
+    #                        - :message: Reports a simple message.
+    #                        - :none or nil: Takes no action.
```
mokagio:
```suggestion
    #                        - Any other value or nil: Takes no action.
```

lib/dangermattic/plugins/milestone_checker.rb
```diff
@@ -35,35 +35,27 @@ class MilestoneChecker < Plugin
...
-    # @param if_no_milestone [Symbol] Action to take if the pull request is not assigned to a milestone. Possible values:
-    #                        - :warn (default): Reports a warning.
-    #                        - :error: Reports an error.
-    #                        - :none or nil: Takes no action.
+    # @param report_type [Symbol] (optional) The type of report for the message. Types: :error, :warning (default), :message.
+    # @param report_type_if_no_milestone [Symbol] Action to take if the pull request is not assigned to a milestone. Possible values:
+    #                                 - :warning (default): Reports a warning.
+    #                                 - :error: Reports an error.
+    #                                 - :message: Reports a message.
+    #                                 - :none or nil: Takes no action.
```
mokagio: What do you think of using the same lean structure for both entries?

```suggestion
    # @param report_type [Symbol] (optional) The type of message for when the PR is has passed over the `days_before_due` threshold. Types: :error, :warning (default), :message.
    # @param report_type_if_no_milestone [Symbol] The type of message for when the PR is not assigned to a milestone. Types: :error, :warning (default), :message. You can also pass :none to not leave a message when there is no milestone.
```

lib/dangermattic/plugins/pr_size_checker.rb
```diff
@@ -43,37 +43,31 @@ class PRSizeChecker < Plugin
     # @param type [:insertions, :deletions, :all] The type of diff size to check. (default: :all)
     # @param max_size [Integer] The maximum allowed size for the diff. (default: DEFAULT_MAX_DIFF_SIZE)
     # @param message [String] The message to display if the diff size exceeds the maximum. (default: DEFAULT_DIFF_SIZE_MESSAGE)
-    # @param fail_on_error [Boolean] If true, fail the PR check when the diff size exceeds the maximum (default: false).
+    # @param report_type [Boolean] If true, fail the PR check when the diff size exceeds the maximum (default: false).
```
mokagio:
```suggestion
    # @param report_type [Symbol] (optional) The type of report for the message. Types: :error, :warning (default), :message.
```

mokagio (approved):
Left suggestions for some minor documentation typos. Approving to unblock.

lib/dangermattic/plugins/milestone_checker.rb
iangmaia: I had changed it slightly so it would be similar (before, the "if_no_milestone" would take just `:warn`). But I think the documentation was a bit off and mentioning `:none`. Thanks!

PR #43, opened 2024-02-02
Tracks Checker Improvements

iangmaia:
As a long overdue follow up to https://wp.me/p91TBi-aE9-p2, this PR updates the Tracks Checker plugin.

The changes:
- Removal of the over specific defaults.
- A new `Tracks` label check, instead of just a text instruction.
- Update of the reported message

### DIFF
--- a/lib/dangermattic/plugins/tracks_checker.rb
+++ b/lib/dangermattic/plugins/tracks_checker.rb
@@ -24,6 +24,8 @@ module Danger
       - Please consider registering any new events.
     MESSAGE
 
+    TRACKS_NO_LABEL_MESSAGE = 'Please ensure the PR has the `Tracks` label.'
+
     # Checks the PR diff for changes in Tracks-related files and provides instructions if changes are detected
     #
     # @param tracks_files [Array<String>] List of Tracks-related file names to check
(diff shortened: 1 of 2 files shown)


lib/dangermattic/plugins/tracks_checker.rb
```diff
@@ -15,34 +15,31 @@ module Danger
...
-    ].freeze
-
     TRACKS_PR_INSTRUCTIONS = <<~MESSAGE
-      This PR contains changes to Tracks-related logic. Please ensure the following are completed:
-      **PR Author**
-      - The PR must be assigned the **Tracks** label
-      **PR Reviewer**
+      This PR contains changes to Tracks-related logic. Please ensure (**author and reviewer**) the following are completed:
+
+      - The PR must be assigned the **Tracks** label.
```
mokagio: This hardcoded label requirement gave me a feeling of "unease". I'm okay to ship with it, but I have a suggestion for your consideration.

The previous implementation [had the built-in requirement of using the "Tracks" label](https://github.com/Automattic/peril-settings/blob/224949002327b59f990e2e31e3b9b4c01f199afb/org/pr/check-tracks.ts#L58), so I guess we can live with the plugin having it, too. We're maintaining the status quo. Which is why, again, I think it's okay.

But the reason I feel uneasy is that this goes the opposite way of #44 where some non-documented, non-intuitive defaults were removed.

What do you think to update to take a `tracks_label` as an input and interpolate it in the message printed if a violation is found. It would slightly complicate the code because this wouldn't be a neat constant anymore, but we'd get more flexibility.

The only case I can think of for enforcing "Tracks" at this level, would be if we wanted a consistent approach to how PRs that engage with Tracks are made across the org. I could see it being useful for an hypothetic Tracks team, who might want to load all the open PRs across our GitHub org that have the Tracks label.

mokagio (approved):

lib/dangermattic/plugins/tracks_checker.rb
iangmaia: Good point 👍 

I've tweaked the code a bit also to take into account a nil `tracks_label`. Implemented here: 74905cad85d24223fe5483030f068a12651f2fb8.

mokagio: Thanks @iangmaia 🚀

PR #44, opened 2024-02-02
Remove non intuitive defaults

iangmaia:
This PR was based on a few other discussions we had in the apps' side on whether some of the default values in a few plugins would create "obscure" behavior.
This PR removes these values so that the apps will have to provide them, improving the readability of the check call in the `Dangerfile`s.


mokagio (approved):
Thank you for this. 

Finding the right balance between reasonable defaults and clear APIs is tricky. I'm glad you shipped the code as it was and used the feedback from going through the project setting up `Dangerfile`s to adjust. Better than trying to get it right the first time abstractly.

PR #58, opened 2024-02-16
Check for Tracks only in added / removed diff lines

iangmaia:
It was brought to my attention by @kidinov that the PR https://github.com/woocommerce/woocommerce-android/pull/10816 had the Tracks check flagged but no actual Tracks-related calls or files changed.

Checking [the diff used by Dangermattic](https://patch-diff.githubusercontent.com/raw/woocommerce/woocommerce-android/pull/10816.diff), I noticed that there are some Tracks calls in the context parts of the diff, which shouldn't be considered in the check.
This PR changes the code so that only additions / removals are considered.

### DIFF
--- a/lib/dangermattic/plugins/common/git_utils.rb
+++ b/lib/dangermattic/plugins/common/git_utils.rb
@@ -84,6 +84,7 @@ module Danger
     # @param change_type [Symbol, nil] Change type to filter lines (e.g., :added, :removed) or nil for no filter
     # @return [Array<MatchedData>] Array of MatchedData objects representing matched lines in files
     def matching_lines_in_diff_files(files:, line_matcher:, change_type: nil)
+      change_types = Array(change_type).map(&:to_sym)
       matched_data = []
 
       files.each do |file|
@@ -92,7 +93,7 @@ module Danger
         diff = danger.git.diff_for_file(file)
 
         diff.patch.each_line do |line|
-          matched_lines << line if line_matcher.call(line) && (change_type.nil? || change_type(diff_line: line) == change_type)
+          matched_lines << line if line_matcher.call(line) && (change_type.nil? || change_types.include?(change_type(diff_line: line)))
         end
 
         matched_data << MatchedData.new(file, matched_lines) unless matched_lines.empty?
--- a/lib/dangermattic/plugins/common/git_utils.rb
+++ b/lib/dangermattic/plugins/common/git_utils.rb
@@ -81,7 +81,8 @@ module Danger
     #
     # @param files [Array<String>] List of file names to check
     # @param line_matcher [Proc] A callable that takes a line and returns true if it matches the desired pattern
-    # @param change_type [Symbol, nil] Change type to filter lines (e.g., :added, :removed) or nil for no filter
+    # @param change_type [Symbol, Array<Symbol>, nil] Change type(s) to filter lines (e.g., `:added`, `:removed`,
+    #                    `:context`, `[:added, :removed]`, …), or nil for no filter
     # @return [Array<MatchedData>] Array of MatchedData objects representing matched lines in files
     def matching_lines_in_diff_files(files:, line_matcher:, change_type: nil)
       change_types = Array(change_type).map(&:to_sym)
--- a/lib/dangermattic/plugins/tracks_checker.rb
+++ b/lib/dangermattic/plugins/tracks_checker.rb
@@ -60,24 +60,10 @@ module Danger
     end
 
     def diff_has_tracks_changes?(tracks_usage_matchers:)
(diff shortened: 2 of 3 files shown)


lib/dangermattic/plugins/tracks_checker.rb
```diff
@@ -60,10 +60,24 @@ def changes_tracks_files?(tracks_files:)
...
+
+      tracks_changes_on_additions || tracks_changes_on_removals
+    end
+
+    def diff_has_tracks_changes_for_change_type?(tracks_usage_matchers:, change_type:)
       matched_lines = git_utils.matching_lines_in_diff_files(
         files: git_utils.all_changed_files,
         line_matcher: ->(line) { tracks_usage_matchers.any? { |tracks_usage_match| line.match(tracks_usage_match) } },
-        change_type: nil
+        change_type: change_type
```
AliSoftware: I wonder if it wouldn't be worth it to update `git_utils.matching_lines_in_diff_files` so that its `change_type` parameter could be either a single symbol (like it currently is)… or an array of symbols?

That way you'd be able to revert your changes above in this `tracks_checker.rb` code (having to create a DRY function and call it twice with 2 different `change_type`) and instead would change the trunk code from `change_type: nil` to `change_type: [:added, :removed]`.

Maybe I'm thinking too much ahead, but I feel like asking for both additions and deletions could be useful in contexts and plugins/checks others than the one for Tracks.

---

Besides, the change to make `matching_lines_in_diff_files` accept an array would likely be simpler than the changes you did above after all 😛 :
 - Add `change_types = Array(change_type).map(&:to_sym)` at the start of the method, so that if `change_type` is a single value (String or Symbol) it will be wrapped inside an `Array`; while if it's already an `Array` it won't do anything to it. (and if it's `nil` that will produce an empty Array `[]`)
 - Change your condition in `matched_lines << line if …` from including `change_type(diff_line: line) == change_type` to `change_types.include?(change_type(diff_line: line))`

iangmaia: It actually crossed my mind, but that would break the API -- though I'm 100% certain no one uses this method, so I wouldn't do a 2.0 just for this anyway 😅 . 

Another alternative would be to add a different symbol, like `:added_or_removed`, which I think would cover most cases (it's hard to think of someone filtering for "added or context"), but still it wouldn't be as flexible as an array.

AliSoftware: Would that break the API though ? The solution I proposed would still accept **either** an Array **or** a single item, so I think that's ok?

I mean, that's the whole reason why I specifically suggested that implementation with [`Array(x)`](https://ruby-doc.org/core-2.7.0/Kernel.html#method-i-Array), as using `change_types = Array(change_type)` will guarantee backward-compatibility. You would still be able to pass a single value like `:added` at call site (this keeping backwards-compat), and `change_types = Array(:added)` will turn wrap the input parameter and end up with `change_types = [:added]` for the rest of your implementation to use it as an array of one element; but now you _also_ have to possibility for your caller to pass an array like `[:added, :removed]` and `change_types = Array([:added, :removed]` will set `change_types = [:added, :removed]` without wrapping the parameter when it's already an Array.

iangmaia: Ahh, gotcha, I haven't realized it wouldn't break the API. Thanks for the suggestion, I'll do it! 👍

iangmaia: Updated on 21992a4bed538f2a06e465af8c1e81e5fe245b47

lib/dangermattic/plugins/common/git_utils.rb
```diff
@@ -84,6 +84,7 @@ def check_added_diff_lines(file_selector:, line_matcher:, message:, report_type:
     # @param change_type [Symbol, nil] Change type to filter lines (e.g., :added, :removed) or nil for no filter
```
AliSoftware:
```suggestion
    # @param change_type [Symbol, Array<Symbol>, nil] Change type(s) to filter lines (e.g., `:added`, `:removed`,
    #                    `:context`, `[:added, :removed]`, …), or nil for no filter
```

spec/tracks_checker_spec.rb
```diff
@@ -147,6 +154,17 @@ module Danger
         end
       end
 
+      def allow_diff_match_(modified_files:, diff_line:, expect_match:)
```
AliSoftware: Is it intentional that this method's name ends with `_`? Seems a bit odd-looking 🙃

iangmaia: Thanks, that was probably a typo + `Replace all` 🙃 Fixed on 21992a4bed538f2a06e465af8c1e81e5fe245b47

AliSoftware (approved):

PR #64, opened 2024-04-23
Reusable GHA Workflow for retrying Buildkite jobs

iangmaia:
This PR is the implementation of the idea of using GHA to trigger a job retry on the Buildkite, as opposed to running the entirety of a Danger job on GHA.
This becomes more important now to leverage the new Linter Agents currently running on Buildkite.

A PR using the present reusable workflow can be seen here: https://github.com/Automattic/pocket-casts-ios/pull/1630

See the original thread on paaHJt-5Qn-p2 for more details.

### DIFF
--- a/.github/workflows/reusable-retry-buildkite-step-on-events.yml
+++ b/.github/workflows/reusable-retry-buildkite-step-on-events.yml
@@ -1,0 +1,110 @@
+on:
+  workflow_call:
+    inputs:
+      org-slug:
+        required: true
+        type: string
+      pipeline-slug:
+        required: true
+        type: string
+      retry-step-key:
+        required: true
+        type: string
+      commit-sha:
+        required: true
+        type: string
+    secrets:
+      buildkite-api-token:
+        required: true
+
+jobs:
+  retry-buildkite-job:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "🔄 Retry job on the latest Buildkite Build"
+        run: |
+          ORG_SLUG="${{ inputs.org-slug }}"
+          PIPELINE_SLUG="${{ inputs.pipeline-slug }}"
+          RETRY_STEP_KEY="${{ inputs.retry-step-key }}"
+          BUILDKITE_API_ACCESS_TOKEN="${{ secrets.buildkite-api-token }}"
+          COMMIT_SHA="${{ inputs.commit-sha }}"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-retry-buildkite-step-on-events.yml
```diff
@@ -0,0 +1,123 @@
...
+              curl \
+                --fail-with-body \
+                --silent \
+                --show-error \
+                -X "$METHOD" \
+                -H "Authorization: Bearer $BUILDKITE_API_ACCESS_TOKEN" \
+                "$BUILDKITE_API_URL"
+            )
+
+            echo "$RAW_RESPONSE" | tr -d '\n' | jq -R -r
```
AliSoftware: What's the goal of using `-R -r` here? 🤔

AliSoftware: Also wouldn't the `tr` command risk removing newlines also within strings which are part of the response (thus modifying the response content itself, not just its formatting)? (e.g. `{ "BUILDKITE_MESSAGE": "Hello\nWorld" }`)

iangmaia: Thanks for pointing those out -- tbh, I can't remember why I originally had those 🙃. Perhaps the `tr` was helpful when developing?
They didn't seem to break the result but you're right, they're not necessary. I've fixed a few other shellcheck issues together with these on ec6dca9257a68818aeaf82741bff0d4f09f577a0.

AliSoftware (approved):
Just checked the code and looks good to me, but I don't have the bandwidth to test it myself in practice on client repos.

Since you mentioned you tested it on a PCiOS PR I trust that it works as expected and will approve to unblock; but feel free to ping another pair of eyes if you want this to be tested more thoroughly.

iangmaia:
> Just checked the code and looks good to me, but I don't have the bandwidth to test it myself in practice on client repos.
> 
> Since you mentioned you tested it on a PCiOS PR I trust that it works as expected and will approve to unblock; but feel free to ping another pair of eyes if you want this to be tested more thoroughly.

Thanks!
I'll merge it now to focus on it from the clients perspective, so they can use what's on trunk. I have a few related draft PRs using this workflow and it seems fine so far, so we can start from there.

PR #76, opened 2024-08-05
Bump Ruby dependencies

AliSoftware:
aka running `bundle update`, as a follow-up to https://github.com/Automattic/dangermattic/pull/75 that was opened by `dependabot`.

---

cc @iangmaia not sure what's next after this PR lands, and couldn't find a "Release / Deployment" paragraph in [the README](https://github.com/Automattic/dangermattic/blob/trunk/README.md) either:

 - Should we release a new version of this gem after this PR lands? If so, what's the process? (Update the `CHANGELOG.md` in a PR, then do a GitHub Release + tag and let CI do the `gem push`, I guess… or is there more to it? Would be nice to provide a small bullet-point list in the README)
 - Do we need to update other places to pick up the new release? In particular, I'm thinking of updating the version of `dangermattic` installed in the linter agent on our CI (update the ansible playbook to pick the new version, then re-deploy them?). Again, would be nice to include in the README as well (even though that part is more specific to a8c itself and what we use `dangermattc` for at a8c, and not relevant to Open-Source documentation, while this repo is public and OSS)


AliSoftware:
Interestingly, the update of the `rubocop` gem made it pick up 2 new linter errors on the codebase. @iangmaia can I let you look into those?

iangmaia:
> cc @iangmaia not sure what's next after this PR lands, and couldn't find a "Release / Deployment" paragraph in [the README](https://github.com/Automattic/dangermattic/blob/trunk/README.md) either:
> 
> * Should we release a new version of this gem after this PR lands? If so, what's the process? (Update the `CHANGELOG.md` in a PR, then do a GitHub Release + tag and let CI do the `gem push`, I guess… or is there more to it? Would be nice to provide a small bullet-point list in the README)

Yep, I think it would be nice to release a new version. Since the first `1.0.0` happened not so long ago and I was taking care of it mostly alone, I didn't spend a lot of time documenting the release, but it's good to do now. Overall it shouldn't be very different than [release-toolkit](https://github.com/wordpress-mobile/release-toolkit) (so: `CHANGELOG.md` update; Github Release creating a tag; CI should do the `gem push`). 
Also, a description of the process + a `rake new_release` target would indeed be nice, I'll add it 👍 

Actually I noticed I hadn't published older Github release versions, just did that now.

> * Do we need to update other places to pick up the new release? In particular, I'm thinking of updating the version of `dangermattic` installed in the linter agent on our CI (update the ansible playbook to pick the new version, then re-deploy them?).

Other repos and the Linter Agent (note that I have a [pending PR](https://github.com/Automattic/buildkite-ci/pull/487) for it, the update could perhaps go together with it) will need to update. Also, remember that the apps using a new version is actually more important, as that will be the version installed by the Linter Agent.

> Interestingly, the update of the rubocop gem made it pick up 2 new linter errors on the codebase. @iangmaia can I let you look into those?

Yeah, seem like new checks compared to the previous version? I've fixed them and pushed to this branch.

iangmaia (approved):

AliSoftware:
Thanks 👍 

I'll let you do the merge of the PR yourself, so you can sync it with follow-up steps (new release, etc) 🙂

PR #77, opened 2024-08-05
Create new releases with Rake

iangmaia:
Add new release `Rake` functionality.

I mostly copied over [release-toolkit](https://github.com/wordpress-mobile/release-toolkit)'s `rakelib`.
I considered other alternatives such as creating a separate Gem (seemed like a bigger effort for not a lot of code) or using a Submodule (seemed odd to have a dependency from Dangermattic to release-toolkit), so just copying the code to start with seemed like an ok trade-off for now.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -11,7 +11,7 @@ task all: %i[specs lint]
 
 desc 'Ensure that the plugin passes `danger plugins lint`'
 task :danger_lint do
-  sh 'bundle exec danger plugins lint'
+  sh('bundle', 'exec', 'danger', 'plugins', 'lint')
 end
 
 desc 'Runs linting tasks: :rubocop and :danger_lint'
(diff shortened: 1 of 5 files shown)


Rakefile
```diff
@@ -22,3 +22,90 @@ RSpec::Core::RakeTask.new(:specs)
 
 desc 'Run RuboCop on the lib/specs directory'
 RuboCop::RakeTask.new(:rubocop)
+
+desc 'Generate the docs using YARD'
+task :doc do
+  sh('yard', 'doc')
+  # Open generated doc in browser
+  sh('open', 'yard-doc/index.html')
```
AliSoftware, line 30: This is not the default location where the documentation is generated iinm. This is the location in `release-toolkit` because that output dir is set [via the `.yardopts` file](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/.yardopts)

AliSoftware (approved):
I agree that at some point it would be nice to make a gem for at least the `ChangelogParser` class, ideally even improving its implementation for it to be less bare-bones but instead with a nicer API to use and probably based on existing gems to parse CommonMark / Markdown and print stuff to Console, instead of re-inventing the wheel on those 2 aspects from scratch. I'm sure there are a couple of tools of ours like `release-toolkit`, `dangermattic`, `hostmgr` etc… that could benefit from DRY-ing all that.

But also agreed that we can probably keep that for a separate request (maybe add an item in our roadmap for it?).

Rakefile
iangmaia: As I've tested it and it worked, so I didn't realize it. But it turns out Dangermattic also [has a .yardopts file](https://github.com/Automattic/dangermattic/blob/trunk/.yardopts).

AliSoftware: 🤦 I thought I checked that but apparently I missed this file in there too 😅 👍

PR #82, opened 2024-09-17
Add cancel parameter to Issue Validation workflow

iangmaia:
This PR adds the parameter `cancel-running-jobs` to the Issue Validation workflow, so that users can choose not to cancel currently running jobs.

### DIFF
--- a/.github/workflows/reusable-check-labels-on-issues.yml
+++ b/.github/workflows/reusable-check-labels-on-issues.yml
@@ -28,7 +28,7 @@ on:
         required: true
 
 concurrency:
-  group: danger-${{ github.event.issue.number }}
+  group: ${{ github.workflow }}-${{ github.event.issue.number }}
   cancel-in-progress: ${{ inputs.cancel-running-jobs }}
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-check-labels-on-issues.yml
```diff
@@ -18,13 +18,18 @@ on:
...
+        type: boolean
+        required: false
     secrets:
       github-token:
         required: true
 
 concurrency:
-  group: danger-${{ github.event.issue.number }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event.issue.number }}
```
mokagio, line 31: Using `${{ github.workflow }}` is a neat tidying.

For my learning, I assume it's is just a tidy up, though, or was it required for the change to work as expected?

mokagio (approved):
Looking good from my understanding.

.github/workflows/reusable-check-labels-on-issues.yml
iangmaia: Yep, it shouldn't have any impact. It's just to make it clearer (and perhaps safer) that this grouping should be unique per workflow.

PR #86, opened 2024-10-28
Add check for `Package.swift` / `Package.resolved` using full paths

iangmaia:
This PR introduces the method `check_swift_package_resolved_updated_strict` to the `ManifestPRChecker` plugin to validate Swift Package Manager manifest changes, now considering a full path:

```ruby
manifest_pr_checker.check_swift_package_resolved_updated_strict(
  manifest_path: "Modules/Package.swift",
  manifest_lock_path: "MyApp.xcworkspace/xcshareddata/swiftpm/Package.resolved"
)
```

I've considered keeping a single method but ended up splitting them to make things clearer, though I'm open to feedback and ideas.

Additionally, I've also changed the plugin to consider in Manifest checks ALL modified files, including added / removed ones.

### Testing
I've covered the method with unit tests, so they should succeed and CI should be 🟢 

You can also run this check by pointing directly to this branch in a `Gemfile` in an existing project with a `Dangerfile`, then calling the check as in the example above.

### DIFF
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -29,6 +29,7 @@ module Danger
   #
   class ManifestPRChecker < Plugin
     MESSAGE = '`%s` was changed without updating its corresponding `%s`. %s.'
+    SWIFT_INSTRUCTION = 'Please resolve the Swift packages as appropriate to your project setup (e.g. in Xcode or by running `swift package resolve`)'
 
     # Performs all the checks, asserting that changes on `Gemfile`, `Podfile` and `Package.swift` must have corresponding
     # lock file changes.
@@ -79,7 +80,7 @@ module Danger
       check_manifest_lock_updated(
         file_name: 'Package.swift',
         lock_file_name: 'Package.resolved',
-        instruction: 'Please resolve the Swift packages as appropriate to your project setup (e.g. in Xcode or by running `swift package resolve`)',
+        instruction: SWIFT_INSTRUCTION,
         report_type: report_type
       )
     end
@@ -94,7 +95,7 @@ module Danger
       check_manifest_lock_updated_strict(
         manifest_path: manifest_path,
         manifest_lock_path: manifest_lock_path,
-        instruction: 'Please resolve the Swift packages as appropriate to your project setup (e.g. in Xcode or by running `swift package resolve`)',
+        instruction: SWIFT_INSTRUCTION,
         report_type: report_type
       )
     end
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -79,7 +79,7 @@ module Danger
       check_manifest_lock_updated(
         file_name: 'Package.swift',
         lock_file_name: 'Package.resolved',
-        instruction: 'Please resolve the Swift packages in Xcode',
+        instruction: 'Please resolve the Swift packages as appropriate to your project setup (e.g. in Xcode or by running `swift package resolve`)',
         report_type: report_type
       )
     end
@@ -94,7 +94,7 @@ module Danger
       check_manifest_lock_updated_strict(
         manifest_path: manifest_path,
         manifest_lock_path: manifest_lock_path,
-        instruction: 'Please resolve the Swift packages in Xcode',
+        instruction: 'Please resolve the Swift packages as appropriate to your project setup (e.g. in Xcode or by running `swift package resolve`)',
         report_type: report_type
       )
     end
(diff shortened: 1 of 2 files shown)


lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -87,20 +84,46 @@ def check_swift_package_resolved_updated(report_type: :warning)
...
+    # checking for exact path matches
+    #
+    # @param report_type [Symbol] (optional) The type of report for the message. Types: :error, :warning (default), :message.
+    #
+    # @return [void]
+    def check_swift_package_resolved_updated_strict(manifest_path:, manifest_lock_path:, report_type: :warning)
+      check_manifest_lock_updated_strict(
+        manifest_path: manifest_path,
+        manifest_lock_path: manifest_lock_path,
+        instruction: 'Please resolve the Swift packages in Xcode',
```
AliSoftware: If one passes `manifest_path: 'Package.swift', manifest_lock_path: 'Package.resolved'` because they are not using an Xcode project but only a bare Swift package (e.g. `Gravatar-SDK-iOS`), then it doesn't make sense to have the instruction be "Please resolve the Swift packages in Xcode" 🙃

iangmaia: Ah, yes 🤦‍♂️  good catch!

iangmaia: Updated on 5590610a803f86f98648effaea8b4395d2f6ef3c

AliSoftware (approved):
Looks good, so approving to unblock; but I think we could improve the default behavior by adding auto-discovery still.

In particular, with the new method being a separate one for `_strict` mode, client projects won't be able to use `check_all_manifest_lock_updated` to check all the manifests at once if they want to now call `check_swift_package_resolved_updated_strict` instead of just `check_swift_package_resolved_updated`

So maybe in a follow-up PR we could try to auto-discover the list of Swift manifests from globbing of the local files, and make the default `check_swift_package_resulted_updated` call the `check_swift_package_resolved_updated_strict` one automatically with the right list?

I think such auto-discovery logic could be done by using something like `Dir.glob('**/Package.{swift,resolved}).reject { |p| p.include?('.xcworkspace')  }`.
This is because, if I'm not mistaken, this Danger plugin can only detect changes in `Package.resolved` that are associated with a `Package.swift` file used by `swift package`, while `Package.resolved` files that are within a `.xcworkspace/*` folder are for package dependencies declared in the corresponding `.pbxproj`, and this Danger plugin doesn't go as far as to decoding and analyzing the content of those `.pbxproj` files for the specific `XCPackageDependency` nodes or whatever to check if those changed and compare with a corresponding `.xcworkspace/**/Package.resolved` change.

So that means that `check_swift_package_resolved_updated*` can only do such a check if there's a `Package.swift` associated with it, managed by `swift package`, not for `Package.resolved` managed by a `pbxproj`. Making the auto-discovery logic not that complicated to implement after all?

iangmaia:
> Looks good, so approving to unblock; but I think we could improve the default behavior by adding auto-discovery still.
> 
> In particular, with the new method being a separate one for `_strict` mode, client projects won't be able to use `check_all_manifest_lock_updated` to check all the manifests at once if they want to now call `check_swift_package_resolved_updated_strict` instead of just `check_swift_package_resolved_updated`
> 
> So maybe in a follow-up PR we could try to auto-discover the list of Swift manifests from globbing of the local files, and make the default `check_swift_package_resulted_updated` call the `check_swift_package_resolved_updated_strict` one automatically with the right list?
> 
> I think such auto-discovery logic could be done by using something like `Dir.glob('**/Package.{swift,resolved}).reject { |p| p.include?('.xcworkspace') }`. This is because, if I'm not mistaken, this Danger plugin can only detect changes in `Package.resolved` that are associated with a `Package.swift` file used by `swift package`, while `Package.resolved` files that are within a `.xcworkspace/*` folder are for package dependencies declared in the corresponding `.pbxproj`, and this Danger plugin doesn't go as far as to decoding and analyzing the content of those `.pbxproj` files for the specific `XCPackageDependency` nodes or whatever to check if those changed and compare with a corresponding `.xcworkspace/**/Package.resolved` change.
> 
> So that means that `check_swift_package_resolved_updated*` can only do such a check if there's a `Package.swift` associated with it, managed by `swift package`, not for `Package.resolved` managed by a `pbxproj`. Making the auto-discovery logic not that complicated to implement after all?

True. For now we'll have to break down the calls, indeed.
My concern about auto discovery relates to similar cases [highlighted in this other Danger check](https://github.com/Automattic/dangermattic/pull/87#discussion_r1823102549) (e.g. we might miss such project specifics also here). For now I'll merge this to have an alternative that can be used but yeah we might be able to make it smarter in a next PR.

PR #87, opened 2024-10-28
Add check for multiple `Package.resolved` files

iangmaia:
This PR adds a new method `check_duplicate_package_resolved` to the `ManifestPRChecker` plugin that helps identify duplicate `Package.resolved` files across SwiftPM root directories.

Example:
```ruby
manifest_pr_checker.check_duplicate_package_resolved(
  spm_root_paths: ['Modules', 'MyApp.xcworkspace']
)
```

### Testing
I've covered the method with unit tests, so they should succeed and CI should be 🟢 

You can also run this check by pointing directly to this branch in a `Gemfile` in an existing project with a `Dangerfile`, then calling the check as in the example above.

### DIFF
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -99,6 +99,23 @@ module Danger
       )
     end
 
+    # Check for duplicate Package.resolved files across all provided SwiftPM package directories
+    #
+    # @param report_type [Symbol] (optional) The type of report for the message. Types: :error, :warning (default), :message.
+    # @param spm_root_paths [Array<String>] Array of paths to SwiftPM package root directories (e.g. ['Modules'])
+    #
+    # @return [void]
+    def check_duplicate_package_resolved(spm_root_paths:, report_type: :warning)
+      resolved_files = spm_root_paths.flat_map { |root_path| Dir.glob("#{root_path}/**/Package.resolved") }
+      return if resolved_files.length <= 1
+
+      reporter.report(
+        message: "Multiple Package.resolved files found:\n#{resolved_files.join("\n")}\n" \
+                 'Please ensure only one Package.resolved exists.',
+        type: report_type
+      )
+    end
+
     private
 
     def check_manifest_lock_updated(file_name:, lock_file_name:, instruction:, report_type: :warning)
--- a/lib/dangermattic/plugins/manifest_pr_checker.rb
+++ b/lib/dangermattic/plugins/manifest_pr_checker.rb
@@ -102,16 +102,16 @@ module Danger
     # Check for duplicate Package.resolved files across all provided SwiftPM package directories
     #
     # @param report_type [Symbol] (optional) The type of report for the message. Types: :error, :warning (default), :message.
-    # @param spm_root_paths [Array<String>] Array of paths to SwiftPM package root directories (e.g. ['Modules'])
     #
     # @return [void]
-    def check_duplicate_package_resolved(spm_root_paths:, report_type: :warning)
-      resolved_files = spm_root_paths.flat_map { |root_path| Dir.glob("#{root_path}/**/Package.resolved") }
-      return if resolved_files.length <= 1
+    def check_swift_duplicate_package_resolved(report_type: :warning)
+      package_files = Dir.glob('./**/Package.swift')
+      resolved_files = Dir.glob('./**/Package.resolved')
+      return if package_files.length == resolved_files.length
 
       reporter.report(
-        message: "Multiple Package.resolved files found:\n#{resolved_files.join("\n")}\n" \
-                 'Please ensure only one Package.resolved exists.',
+        message: "Multiple `Package.resolved` files found:\n```#{resolved_files.join("\n")}```\n" \
+                 'Please ensure only one `Package.resolved` exists for each `Package.swift` file.',
         type: report_type
       )
     end
--- a/spec/manifest_pr_checker_spec.rb
+++ b/spec/manifest_pr_checker_spec.rb
@@ -175,6 +175,39 @@ module Danger
             expect(@dangerfile).to not_report
           end
         end
+
+        describe '#check_duplicate_package_resolved' do
+          it 'reports a warning when multiple Package.resolved files are found' do
+            allow(Dir).to receive(:glob).with('Modules/**/Package.resolved').and_return(['Modules/Package.resolved'])
+            allow(Dir).to receive(:glob).with('MyApp.xcworkspace/**/Package.resolved').and_return(['MyApp.xcworkspace/xcshareddata/swiftpm/Package.resolved'])
+
+            @plugin.check_duplicate_package_resolved(spm_root_paths: ['Modules', 'MyApp.xcworkspace'])
+
+            expected_warning = "Multiple Package.resolved files found:\n" \
+                               "Modules/Package.resolved\n" \
+                               "MyApp.xcworkspace/xcshareddata/swiftpm/Package.resolved\n" \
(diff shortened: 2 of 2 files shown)


lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -99,6 +99,23 @@ def check_swift_package_resolved_updated_strict(manifest_path:, manifest_lock_pa
       )
     end
 
+    # Check for duplicate Package.resolved files across all provided SwiftPM package directories
+    #
+    # @param report_type [Symbol] (optional) The type of report for the message. Types: :error, :warning (default), :message.
+    # @param spm_root_paths [Array<String>] Array of paths to SwiftPM package root directories (e.g. ['Modules'])
```
AliSoftware: I wonder if we couldn't auto-detect those though, by globbing the repo for all `Package.swift` files and `xcodeproj`/`xcworkspace` files 🤔

lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -99,6 +99,23 @@ def check_swift_package_resolved_updated_strict(manifest_path:, manifest_lock_pa
...
+    #
+    # @return [void]
+    def check_duplicate_package_resolved(spm_root_paths:, report_type: :warning)
+      resolved_files = spm_root_paths.flat_map { |root_path| Dir.glob("#{root_path}/**/Package.resolved") }
+      return if resolved_files.length <= 1
```
AliSoftware: If I'm understanding the way some repos use SPM, I think it might still be totally valid to have multiple `Package.resolved` files in the repo, when you have multiple (local) SwiftPM packages in a monorepo

For example, a repo could use:
 - An `xcworkspace` at the root of the repo for the main app project, which might have some (local & 3rd party) packages defined in the "Package Dependencies" tab of its Xcode project
 - But also in order to organize the code of the app in submodules, might also have a `Modules/` or `Components/` subfolder, with potentially many local packages in there (either one per sub-component, or a single package defining all the components in a single `Package.swift`).

In such cases I'd expect as many `Package.resolved` as there are package definitions (either in a `Package.swift` or in dependencies defined direclty in a `xcodeproj`'s "Package Dependencies" tab) in total in the repo…

For example, Tumblr/iOS has:
 - [`Orangina.xcworkspace/xcshareddata/swiftpm/Package.resolved`](https://github.tumblr.net/Tumblr/ios/tree/develop/Orangina.xcworkspace/xcshareddata/swiftpm), which corresponds to the lockfile for the Package Dependencies defined in the Xcode projects from `Orangina.xcworkspace` (see `XCSwiftPackageProductDependency` in `Orangina.xcodeproj/project.pbxproj`)
 - [`Components/Package.resolved`](https://github.tumblr.net/Tumblr/ios/blob/develop/Components/Package.resolved), which corresponds to the lockfile for [`Components/Package.swift`](https://github.tumblr.net/Tumblr/ios/blob/develop/Components/Package.swift)

Such a setup, with two `Package.resolved` in the repo, is totally valid, as long as each package definition source (either `Package.swift` file or `XCSwiftPackageProductDependency` entry in an Xcode project's `pxbproj`) has only one matching `Package.resolved` corresponding to that package.

iangmaia: > If I'm understanding the way some repos use SPM, I think it might still be totally valid to have multiple Package.resolved files in the repo, when you have multiple (local) SwiftPM packages in a monorepo

Yep, that was also my understanding. Isn't the list of separate roots (e.g. `Components` and `Orangina.xcworkspace`) covering that scenario? 🤔

AliSoftware: Not given your current implementation iiuc because you error if the total number of Package.resolved files in the provided paths is > 1 while in those situations it would be expected to be one per provided path, not one total.

So maybe it's just your implementation that is incorrect and should be checking for `== spm_root_paths.length` not on `<= 1`

lib/dangermattic/plugins/manifest_pr_checker.rb
iangmaia: Yeah that was my first thought. I considered it would make it a bit harder to find the `Package.resolved` file(s) corresponding to a given `Package.swift` (though we could also just see if the overall number matches 🤔), but maybe I'm missing something from the possible scenarios with `xcodeproj` / `xcworkspace` 🤔

iangmaia: I'm thinking of going on the route of checking for matching `Package.swift` and `Package.resolved`, without a list of paths 🤔 seems simple enough and should solve the issue.

iangmaia: Pushed another implementation just comparing the count and globbing the repo (removing the list param) at 7ca1da561eb090434f87a891f3b039984d7296dc

lib/dangermattic/plugins/manifest_pr_checker.rb
```diff
@@ -99,6 +99,23 @@ def check_swift_package_resolved_updated_strict(manifest_path:, manifest_lock_pa
...
 
+    # Check for duplicate Package.resolved files across all provided SwiftPM package directories
+    #
+    # @param report_type [Symbol] (optional) The type of report for the message. Types: :error, :warning (default), :message.
+    #
+    # @return [void]
+    def check_swift_duplicate_package_resolved(report_type: :warning)
+      package_files = Dir.glob('./**/Package.swift')
+      resolved_files = Dir.glob('./**/Package.resolved')
+      return if package_files.length == resolved_files.length
```
AliSoftware, line 110: Thinking about it more, this is actually still not the right test, because one can have a `Package.resolved` that is generated by `xcodebuild` within the `.xcworkspace` or `.xcodeproj` (if those package dependencies are managed by Xcode), in which case there will be a `Package.resolved` file within the `.xcworkspace/**` folder without a corresponding `Package.swift` in the repo, and the counts won't match.

I see multiple use cases:
 - Project fully managed by SwiftPM, not `xcodeproj` nor `xcworkspace`: we expect `count(Package.swift) == count(Package.resolved)`
 - Project containing only a `xcodeproj`: we expect only a `Package.resolved` within the `xcodeproj/`, and should report a violation if there's a `Package.resolved` outside the `xcodeproj` (e.g. at the root of the repo) without a corresponding `Package.swift` in the same directory
  - Project containing a `xcworkspace` and one or more `xcodeproj`: not sure if we expect only a `Package.resolved` in the `xcworkspace/**` or if there's a possibiilty that Xcode would _also_ generate a `Package.resolved` in the `.xcodeproj`  bundles that the `xcodeproj` references (especially if each `xcodeproj` might have separate Swift packages declared in the "Package Dependencies" tab of each of the `xcodeproj`…?)

So we should cover all the use cases possible (repo with only SwiftPM and no Xcode; repo with a mix of SwiftPM and Xcode project/workspace for the same product; repo with an Xcode project/workspace at the root + one or more `Package.swift` / `Package.resolved` pairs to develop local packages as a way to modularize development of the product; repo with only Xcode project/workspace without any SPM at all; …)

iangmaia: > Thinking about it more, this is actually still not the right test, because one can have a Package.resolved that is generated by xcodebuild within the .xcworkspace or .xcodeproj (if those package dependencies are managed by Xcode), in which case there will be a Package.resolved file within the .xcworkspace/** folder without a corresponding Package.swift in the repo, and the counts won't match.

Yeah, I was realizing that earlier today when I looked at the PR again, but didn't get back to it 😕 I'll give it another try.

PR #92, opened 2024-11-06
Ignore private classes in unit test checker

### LINKED ISSUE #72: Private classes in public files trigger missing tests check
When a class is private but it is in a public file it triggers the check like [here](https://buildkite.com/automattic/pocket-casts-android/builds/7172#019081b3-557f-44a2-abaf-b4268b66784c). We don't use the latest `1.2.0` version in Pocket Casts but I don't see any changes in the git history that might got it fixed.

Here's a sample of offending code. I'm guessing a bit because I didn't check `dangermattic` project out to test it but it replicates more or less the code from [the failing PR](https://github.com/Automattic/pocket-casts-android/pull/2447).

Assume `Hello.kt` file.

```kotlin
class Hello {
    private val worlds = listOf(Earth(), Mars())

    fun print() = worlds.joinToString(prefix = "Hello, ", separator = ", ", postfix = "!", transform = World::print)
}

private interface World {
    fun print(): String
}

private class Earth : World {
    override fun print() = "Earth"
}

private class Mars : World {
    override fun print() = "Mars"
}
```

MiSikora:
This PR updates the regex used to detect non-private classes. The previous regex was flawed as it captured classes like `private class Foo` due to the `\s` expression in the first non-capturing group. The new regex captures the same data as before but now includes class modifiers and filters based on them.

Closes #72

### DIFF
--- a/lib/dangermattic/plugins/android_unit_test_checker.rb
+++ b/lib/dangermattic/plugins/android_unit_test_checker.rb
@@ -24,7 +24,14 @@ module Danger
   #
   class AndroidUnitTestChecker < Plugin
     ANY_CLASS_DETECTOR = /class\s+([A-Z]\w+)\s*(.*?)\s*{/m
-    NON_PRIVATE_CLASS_DETECTOR = /(?:\s|public|internal|protected|final|abstract|static)*class\s+([A-Z]\w+)\s*(.*?)\s*{/m
+    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum)*)class\s+([A-Z]\w+)\s*(.*?)\s*{/m
+
+    CLASS_MODIFIER_EXCEPTIONS = [
+      /\s*data\s*/,
+      /\s*private\s*/,
+      /\s*enum\s*/,
+    ].freeze
+
     DEFAULT_CLASSES_EXCEPTIONS = [
       /ViewHolder$/,
       /Module$/,
@@ -142,7 +149,7 @@ module Danger
     # @return [Array<ClassViolation>] An array of ClassViolation objects representing the violations found.
     def find_violations(path:, diff_patch:, classes_exceptions:, subclasses_exceptions:)
       added_lines = git_utils.added_lines(diff_patch: diff_patch)
-      matches = added_lines.scan(NON_PRIVATE_CLASS_DETECTOR)
+      matches = added_lines.scan(CLASS_MODIFIER_DETECTOR)
       matches.reject! do |m|
         class_match_is_exception?(
           m,
@@ -152,7 +159,7 @@ module Danger
         )
       end
 
-      matches.map { |m| ClassViolation.new(m[0], path) }
+      matches.map { |m| ClassViolation.new(m[1], path) }
     end
 
     # Finds the names of removed classes based on the removals the diff patch.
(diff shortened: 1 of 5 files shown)


lib/dangermattic/plugins/android_unit_test_checker.rb
```diff
@@ -24,7 +24,14 @@ module Danger
   #
   class AndroidUnitTestChecker < Plugin
     ANY_CLASS_DETECTOR = /class\s+([A-Z]\w+)\s*(.*?)\s*{/m
-    NON_PRIVATE_CLASS_DETECTOR = /(?:\s|public|internal|protected|final|abstract|static)*class\s+([A-Z]\w+)\s*(.*?)\s*{/m
+    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum)*)class\s+([A-Z]\w+)\s*(.*?)\s*{/m
```
iangmaia, line 27: Hey @MiSikora , thanks a lot for the PR 🙂 

The goal of the original regex was to start with a list of eligible classes for the check (thus `NON_PRIVATE_CLASS_DETECTOR`) as a starting point -- that's why it didn't include all possible class modifiers.

IMO I think we could keep that part as it was and add your new `CLASS_MODIFIER_EXCEPTIONS` to exclude exceptions we iterate through the initial matches.

MiSikora: But `NON_PRIVATE_CLASS_DETECTOR` doesn't work as intended. It detects private classes.

spec/android_unit_test_checker_spec.rb
```diff
@@ -334,7 +348,7 @@ def prefix_text_lines(content:, prefix:)
     end
 
     def expect_class_names_match_report(class_names:, error_report:)
-      expect(class_names.length).to eq(error_report.length)
+      # expect(class_names.length).to eq(error_report.length)
```
iangmaia: Hmm why is this check failing? 🤔

MiSikora: Oh sorry, it was from me testing things. I'll revert this change.

MiSikora: Fixed in https://github.com/Automattic/dangermattic/pull/92/commits/50ba674e7d09e771de12ff5eb0467247ffbbb14c

lib/dangermattic/plugins/android_unit_test_checker.rb
iangmaia: Ah, it's taking me some time to get back to this 😅 gotcha. Having a second look at this, it's making more sense (also semantically to start with a complete list that we then filter 🤔).
I'll run a few tests but the code is looking good. Did you test it in a real PR where you had these false positives?

MiSikora: > I'll run a few tests but the code is looking good. Did you test it in a real PR where you had these false positives?

No, I didn't. I'll make a sample PR and get back to you.

MiSikora: As we discussed in Slack, tested with this PR: https://github.com/Automattic/pocket-casts-android/pull/3184.

iangmaia (approved):
Thanks a lot for fixing this long standing bug! 👏

PR #94, opened 2024-11-22
Add `sealed` and `value` classes as an exception when reporting missing Android unit tests

iangmaia:
I figured `sealed` classes, which are often used to model states similarly to `enum`s, should also be an exception when reporting missing unit tests, so this PR adds them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- Add sealed classes as an exception when reporting missing Android unit tests [#94]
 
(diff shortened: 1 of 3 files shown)


MiSikora (approved):
While at it it would be good to add `value` classes to the list - https://kotlinlang.org/docs/inline-classes.html.

iangmaia:
> While at it it would be good to add `value` classes to the list - https://kotlinlang.org/docs/inline-classes.html.

Great point -- I actually wasn't aware of value classes at all! I've added them as well, thanks!

PR #101, opened 2025-09-22
`android_unit_test_checker`: add exception for `annotation class`

AliSoftware:
Closes https://linear.app/a8c/issue/AINFRA-1273/dangermattic-do-not-require-tests-for-annotation-classes

## What

Adds an exception for `annotation class` to the `android_unit_test_checker`

## Testing

Tested in [this dummy Draft PR in DOAndroid](https://github.com/bloom/DayOne-Android/pull/6539) by running `danger pr "https://github.com/bloom/DayOne-Android/pull/6539"` on my Terminal (after having set the API token env var), especially after that PR added [this file](https://github.com/bloom/DayOne-Android/pull/6539/files#diff-e862de46717f5ff22703e5b793b9d49a617f1cec6ad0321fff67f8a4d81be6dc):

```
Results:

Errors:
- [ ] Please add tests for class `StravaAppIntegrationHandler` (or add `unit-tests-exemption` label to ignore this).

Warnings:
- [ ] This PR is larger than 500 lines of changes. Please consider splitting it into smaller PRs for easier and faster reviews.

Messages:
- [ ] This PR is still a Draft: some checks will be skipped.
```

This validates that:
 - The `annotation class` added in that dummy Draft PR didn't trigger the Danger warning anymore
 - Dangermattic still detects a true positive of a newly-introduced `StravaAppIntegrationHandler` class missing tests (as [I intentionally removed them from that dummy draft PR](https://github.com/bloom/DayOne-Android/pull/6539/commits/959560f3593084065d677ca33a62541a8a35fe0d) to validate exactly that scenario)

🟢  Since I also added a corresponding unit test and those are run on CI, CI being green should also be enough for considering this behaves as expected.

### DIFF
--- a/lib/dangermattic/plugins/android_unit_test_checker.rb
+++ b/lib/dangermattic/plugins/android_unit_test_checker.rb
@@ -24,7 +24,7 @@ module Danger
   #
   class AndroidUnitTestChecker < Plugin
     ANY_CLASS_DETECTOR = /class\s+([A-Z]\w+)\s*(.*?)\s*{/m
-    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum|sealed|value|annotation)*)class\s+([A-Z]\w+)\s*(.*?)\s*{/m
+    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum|sealed|value|annotation)*)class\s+([A-Z]\w+)\s*(.*?)\s*(?:{|\n\n)/m
 
     CLASS_MODIFIER_EXCEPTIONS = [
       /\s*data\s*/,
--- a/lib/dangermattic/plugins/android_unit_test_checker.rb
+++ b/lib/dangermattic/plugins/android_unit_test_checker.rb
@@ -153,7 +153,7 @@ module Danger
     # @return [Array<ClassViolation>] An array of ClassViolation objects representing the violations found.
     def find_violations(path:, diff_patch:, classes_exceptions:, subclasses_exceptions:)
       added_lines = git_utils.added_lines(diff_patch: diff_patch)
-      matches = added_lines.scan(CLASS_MODIFIER_DETECTOR)
+      matches = "#{added_lines}\n".scan(CLASS_MODIFIER_DETECTOR) # add a newline to ensure the regex matches the last class in the file
       matches.reject! do |m|
         class_match_is_exception?(
           m,
--- a/lib/dangermattic/plugins/android_unit_test_checker.rb
+++ b/lib/dangermattic/plugins/android_unit_test_checker.rb
@@ -24,7 +24,7 @@ module Danger
   #
   class AndroidUnitTestChecker < Plugin
     ANY_CLASS_DETECTOR = /class\s+([A-Z]\w+)\s*(.*?)\s*{/m
-    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum|sealed|value|annotation)*)class\s+([A-Z]\w+)\s*(.*?)\s*(?:{|\n\n)/m
+    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum|sealed|value|annotation)*)class\s+([A-Z]\w+)\s*(.*?)\s*({|\n\n)/m
 
     CLASS_MODIFIER_EXCEPTIONS = [
       /\s*data\s*/,
@@ -186,6 +186,7 @@ module Danger
     def class_match_is_exception?(match, file, classes_exceptions, subclasses_exceptions)
       return true if classes_exceptions.any? { |re| match[1] =~ re }
       return true if CLASS_MODIFIER_EXCEPTIONS.any? { |re| match[0] =~ re }
+      return true unless match[3].include?('{') # Ignore classes that don't have a body
 
       subclass_regexp = File.extname(file) == '.java' ? /extends\s+([A-Z]\w+)/m : /\s*:\s*([A-Z]\w+)/m
       subclass = match[2].scan(subclass_regexp)&.last&.last
--- a/lib/dangermattic/plugins/android_unit_test_checker.rb
+++ b/lib/dangermattic/plugins/android_unit_test_checker.rb
@@ -182,7 +182,7 @@ module Danger
     # @param classes_exceptions [Array<String>] Regexes matching class names to exclude from the check.
     # @param subclasses_exceptions [Array<String>] Regexes matching base class names to exclude from the check
     #
-    # @return [void]
+    # @return [Boolean]
     def class_match_is_exception?(match, file, classes_exceptions, subclasses_exceptions)
       return true if classes_exceptions.any? { |re| match[1] =~ re }
       return true if CLASS_MODIFIER_EXCEPTIONS.any? { |re| match[0] =~ re }
--- a/lib/dangermattic/plugins/android_unit_test_checker.rb
+++ b/lib/dangermattic/plugins/android_unit_test_checker.rb
@@ -24,14 +24,15 @@ module Danger
   #
   class AndroidUnitTestChecker < Plugin
     ANY_CLASS_DETECTOR = /class\s+([A-Z]\w+)\s*(.*?)\s*{/m
-    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum|sealed|value)*)class\s+([A-Z]\w+)\s*(.*?)\s*{/m
+    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum|sealed|value|annotation)*)class\s+([A-Z]\w+)\s*(.*?)\s*{/m
 
     CLASS_MODIFIER_EXCEPTIONS = [
       /\s*data\s*/,
       /\s*private\s*/,
       /\s*enum\s*/,
       /\s*sealed\s*/,
-      /\s*value\s*/
+      /\s*value\s*/,
+      /\s*annotation\s*/
     ].freeze
 
     DEFAULT_CLASSES_EXCEPTIONS = [
--- a/spec/fixtures/android_unit_test_checker/MixDataPlusNormalClass.kt
+++ b/spec/fixtures/android_unit_test_checker/MixDataPlusNormalClass.kt
@@ -1,0 +1,14 @@
+import javax.inject.Inject
+import javax.inject.Singleton
+
+internal data class DummyDataClassNotNeedingTests(
+    val title: String,
+    val id: String
+)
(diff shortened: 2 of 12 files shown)


lib/dangermattic/plugins/android_unit_test_checker.rb
```diff
@@ -24,14 +24,15 @@ module Danger
   #
   class AndroidUnitTestChecker < Plugin
     ANY_CLASS_DETECTOR = /class\s+([A-Z]\w+)\s*(.*?)\s*{/m
-    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum|sealed|value)*)class\s+([A-Z]\w+)\s*(.*?)\s*{/m
+    CLASS_MODIFIER_DETECTOR = /((?:\s|public|internal|protected|private|final|abstract|static|data|enum|sealed|value|annotation)*)class\s+([A-Z]\w+)\s*(.*?)\s*(?:{|\n\n)/m
```
AliSoftware: ℹ️  Note about the change from `{` to `(?:{|\n\n)` at the end of the RegEx:

This was not _directly_ related to this PR but something I noticed while I was testing that change on https://github.com/bloom/DayOne-Android/pull/6539. 

Indeed, in the file I used for testing the RegEx would still find true positives correctly, I noticed that if the file contains a `data class` with no `{…}` body before another `class` declaration, the `(.*?)` part of the RegEx would match the whole text up until the `{` of the **second** class when matching the first `data class`. Which means that instead of that `(.*?)` capturing the rest of the class signature—including parent class inheritance etc but before the class' body—as intended, it would instead also include the signature of the next class in that file so it can match the next `{` it could find.

This is illustrated by [the test fixture I added below](https://github.com/Automattic/dangermattic/pull/101/files#diff-368583a8827ec84ce1a7d378aa13dc310c8861c3f38747ee12ad537a31f0a68a), where without that change the RegEx would include all of lines 4 to 12 when matching `internal data class DummyDataClassNotNeedingTests`, thus accidentally including the `internal class DummyClassMissingTest` and its `: ParentClass {` as if that was all part of the first `data class`' signature too… and more importantly, that'd mean that the RegEx would then NOT detect the `class DummyClassMissingTest` as a 2nd match and 2nd class to be detected (because part of its declaration was already part of the first match)

This change in the RegEx allows to consider an empty newline at the end of a class declaration as being the end of the class signature too, as an alternative signal to encountering `{` signaling the start of a class body.

spec/fixtures/android_unit_test_checker/MixDataPlusNormalClass.kt
AliSoftware, line 1: 👋  @ParaskP7 I know you might not be super familiar with Dangermattic's codebase and not the best suited for reviewing this PR, but since @iangmaia is on support rotation this week, I figured I'd add you anyway in the meantime so we can merge this before next week.

Since I've already tested this locally (see PR description) I guess your job as reviewer is to validate that:
 - This test fixture example would make sense in Kotlin
 - And that [the test only reporting `DummyClassMisingTest` as a violation](https://github.com/Automattic/dangermattic/pull/101/files#diff-81297f5e382c308b4edcc205e8977872aaf1a3e5e51bf3194aafa541550a7272R159) (but not reporting `DummyDataClassNotNeedingTest` as a violation because it's a `data class`) is the behavior an Android developer would expect from that Danger rule 🙂

ParaskP7: Thanks for the ping and giving me the extra context @AliSoftware , I'll review this first thing tomorrow! 💯

ParaskP7 (approved):
👋 @AliSoftware and thanks for adding this exception for `annotation` classes, it sure will be used to all projects, not only DOAndroid! 💯 ✅ 🚀 

-----

As for [this](https://github.com/Automattic/dangermattic/pull/101#discussion_r2368609311) comment and your questions:

> This test fixture example would make sense in Kotlin

It does! 💯 

> And that [the test only reporting DummyClassMisingTest as a violation](https://github.com/Automattic/dangermattic/pull/101/files#diff-81297f5e382c308b4edcc205e8977872aaf1a3e5e51bf3194aafa541550a7272R159) (but not reporting DummyDataClassNotNeedingTest as a violation because it's a data class) is the behavior an Android developer would expect from that Danger rule 🙂

Yes, indeed! 💯

---

Suggestion (💡): The only bit I would maybe suggest, especially since this PR is all about `annotation` related classes, maybe we want to add such a test fixture example as well. Btw, I am not sure if we have such a test fixture per all other exceptions, like `value`, `sealed`, `enum`, etc. If not, maybe worth adding those as well... 💭

AliSoftware:
Thanks for the review!

> maybe we want to add such a test fixture example as well

I'm not sure I follow what you mean 😅  Isn't [this `MixDataPlusNormalClass.kt` fixture](https://github.com/Automattic/dangermattic/pull/101/files/4a26ce15c33dd8a21eb97be9b1b80a6a7aaa36b9#diff-368583a8827ec84ce1a7d378aa13dc310c8861c3f38747ee12ad537a31f0a68a) already doing exactly that?

> I am not sure if we have such a test fixture per all other exceptions, like value, sealed, enum, etc. If not, maybe worth adding those as well...

Good point, I don't think we have such fixtures so far, worth adding indeed 👍 
(If you have dummy examples of such class types in Kotlin I can copy/paste as fixtures, I'm interested, as not being a Kotlin dev myself I don't know how to write those; I'll ask Claude otherwise in the meantime)

ParaskP7:
Thanks for the quick reply @AliSoftware ! 🙏

> I'm not sure I follow what you mean 😅 Isn't [this MixDataPlusNormalClass.kt fixture](https://github.com/Automattic/dangermattic/pull/101/files/4a26ce15c33dd8a21eb97be9b1b80a6a7aaa36b9#diff-368583a8827ec84ce1a7d378aa13dc310c8861c3f38747ee12ad537a31f0a68a) already doing exactly that?

This text fixture is all about:

1. `data class`
2. `class`

Not about `annotation class`, which this PR tries to solve (see [AppIntegrationHandlers](https://github.com/bloom/DayOne-Android/blob/trunk/app/src/main/java/com/dayoneapp/dayone/main/settings/integrations/connect/AppIntegrationModule.kt#L15) on DOAndroid).

What am I missing? 😊

> Good point, I don't think we have such fixtures so far, worth adding indeed 👍

💯 🙏 

> (If you have dummy examples of such class types in Kotlin I can copy/paste as fixtures, I'm interested, as not being a Kotlin dev myself I don't know how to write those; I'll ask Claude otherwise in the meantime)

For `annotation class` we can just copy-paste this [AppIntegrationHandlers](https://github.com/bloom/DayOne-Android/blob/trunk/app/src/main/java/com/dayoneapp/dayone/main/settings/integrations/connect/AppIntegrationModule.kt#L15) from DOAndroid.

For the rest of them, you can simply do exactly that, just search their project in GitHub for any of the following and pick a random example:

1. [annotation class](https://github.com/search?q=repo%3Abloom%2FDayOne-Android+annotation+class&type=code)
2. [value class](https://github.com/search?q=repo%3Abloom%2FDayOne-Android+value+class&type=code)
3. [sealed class](https://github.com/search?q=repo%3Abloom%2FDayOne-Android%20sealed%20class&type=code)
4. [enum class](https://github.com/search?q=repo%3Abloom%2FDayOne-Android+enum+class&type=code)
5. [private class](https://github.com/search?q=repo%3Abloom%2FDayOne-Android+private+class&type=code)

Hope that helps you? 🙏

AliSoftware:
> What am I missing? 😊

🤦 I am the one missing something (my morning coffee, probably 😅 )

Since I did some manual tests on the [dummy DOAndroid PR](https://github.com/bloom/DayOne-Android/pull/6539) yesterday to test the case of `annotation class` I somehow still had in my mind that the fixture I added already had that `annotation class` too, while I confused it with what I tested in the dummy PR…

Will add all those fixtures today and ping you back for reviewing them once ready 🙇

ParaskP7:
> 🤦 I am the one missing something (my morning coffee, probably 😅 )

☕ x ☕ ^ ☕ 

> Will add all those fixtures today and ping you back for reviewing them once ready 🙇

Awesome, thanks @AliSoftware , I'll be here waiting! ❤️

AliSoftware:
@ParaskP7 took longer than expected to write the tests properly, also because I wanted to also test that we correctly detected all classes at first (even the ones we then exclude due to exceptions on the modifiers like `annotation` / `data`, `enum`…), to validate the RegEx matched all of them… but I finally added all those unit tests 😅 

I'm happy I took the time to do so though, because that revealed a couple of cases that the RegEx was previously missing, especially around classes with no `{…}` body that were the last one at the end of a file (tricky!). So thanks for pushing me to add those tests! 🎉 

The PR is now ready for re-review, especially for you to validate the fixtures I added would cover the main cases and Kotlin syntax variants we'd want to validate for class declarations 🙇

lib/dangermattic/plugins/android_unit_test_checker.rb
```diff
@@ -152,7 +153,7 @@ def find_classes_missing_tests(git_diff:, classes_exceptions:, subclasses_except
     # @return [Array<ClassViolation>] An array of ClassViolation objects representing the violations found.
     def find_violations(path:, diff_patch:, classes_exceptions:, subclasses_exceptions:)
       added_lines = git_utils.added_lines(diff_patch: diff_patch)
-      matches = added_lines.scan(CLASS_MODIFIER_DETECTOR)
+      matches = "#{added_lines}\n".scan(CLASS_MODIFIER_DETECTOR) # add a newline to ensure the regex matches the last class in the file
```
ParaskP7, line 156: Praise (❤️): Nice!

lib/dangermattic/plugins/android_unit_test_checker.rb
```diff
@@ -181,10 +182,11 @@ def find_removed_classes(diff_patch:)
     # @param classes_exceptions [Array<String>] Regexes matching class names to exclude from the check.
     # @param subclasses_exceptions [Array<String>] Regexes matching base class names to exclude from the check
     #
-    # @return [void]
+    # @return [Boolean]
     def class_match_is_exception?(match, file, classes_exceptions, subclasses_exceptions)
       return true if classes_exceptions.any? { |re| match[1] =~ re }
       return true if CLASS_MODIFIER_EXCEPTIONS.any? { |re| match[0] =~ re }
+      return true unless match[3].include?('{') # Ignore classes that don't have a body
```
ParaskP7, line 189: Praise (❤️): Awesome! 🥇 WIN 🥇

ParaskP7 (approved):
This is awesome @AliSoftware thanks for adding and verifying all those (test) cases, this looks so much better now, and the confidence level that the regex works as expected now further increasing! 🙇 ❤️ 🚀 

PS: About the `dayone` classes you added in there, I had a few comments but realized that you just copy-paste them, without caring much, and you did that on purposes, even if some of that code are not too useful for testing, or even duplicate, so I am ignoring that as such, I think it indeed makes sense. 👍

PR #111, opened 2026-03-18
Add LLM review plugin

iangmaia:
Fixes AINFRA-2165

## Summary

Adds a new `llm_reviewer` Danger plugin for automated PR review via LLM APIs.


README.md
```diff
@@ -27,6 +41,15 @@ Once the main Gem is installed, all Dangermattic plugins are available in your `
     # Reports a warning if a pull request diff size is greater than 300
     pr_size_checker.check_diff_size(max_size: 300)
     ```
+- `llm_reviewer` - Performs an automated LLM-based code review on the changed lines in a pull request
+    ```ruby
+    # Reviews the PR diff with OpenAI and posts inline findings
```
mokagio:
```suggestion
    # Reviews the PR diff with GPT-4o and posts inline findings
```

.buildkite/pipeline.yml
```diff
@@ -30,13 +30,9 @@ steps:
...
-    command: |
-      echo "--- :rubygems: Setting up Gems"
-      bundle install
-
-      echo "--- :rubocop: Run Rubocop"
-      bundle exec rubocop
-    plugins: *common_plugins
+    command: rubocop
+    agents:
+      queue: linter
```
mokagio, line 35: Nice

lib/dangermattic/plugins/llm_reviewer.rb
```diff
@@ -0,0 +1,385 @@
...
+
+    SEVERITY_MAP = { 'info' => :message, 'warning' => :warning, 'error' => :error }.freeze
+    TAG_SEVERITY_MAP = { message: 'info', warning: 'warning', error: 'error' }.freeze
+
+    # Pattern to identify and parse LLM review tags in comment bodies
+    LLM_REVIEW_TAG_PATTERN = /<!-- llm-review:(error|warning|info) -->/
+
+    DEFAULT_SYSTEM_PROMPT = <<~PROMPT
+      You are an expert code reviewer. You will receive a pull request diff and metadata.
+      Your job is to identify issues in the CHANGED code (added lines only).
```
mokagio: What about modified lines?

Update: the instructions below mention added and modified lines. With that in mind, I'd suggest removing the parenthetic for clarity:

```suggestion
      Your job is to identify issues in the CHANGED code.
```

lib/dangermattic/plugins/llm_reviewer.rb
```diff
@@ -0,0 +1,385 @@
...
+      capped
+    end
+
+    def report_findings(findings:, default_severity:)
+      findings.each do |finding|
+        severity = SEVERITY_MAP.fetch(finding.severity, default_severity)
+        tag_severity = finding.severity || TAG_SEVERITY_MAP.fetch(severity, 'warning')
+        report_single_finding(finding: finding, severity: severity, tag_severity: tag_severity)
+      end
+    end
```
mokagio, line 365: I don't know how to address this, but the individual reporting approach results in one review post per report

Image

That looks noisy to me.

It might be a Danger limitation, or it might be something we can change to make the report be a standard comment instead of a review comment. Anyway, not a blocker.

README.md
```diff
@@ -8,6 +8,20 @@ Add to your project's `Gemfile`
 gem 'danger-dangermattic', git: 'https://github.com/Automattic/dangermattic'
 ```
 
+### LLM-backed plugin setup
+
+Plugins that call LLM APIs need the matching provider key available in CI:
+
+- `OPENAI_API_KEY` for OpenAI models such as `gpt-4o`
+- `ANTHROPIC_API_KEY` for Anthropic models such as `claude-sonnet-4-20250514`
+
```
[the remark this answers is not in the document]
mokagio: This makes sense to me as an edit. What do you think @iangmaia ?

mokagio (approved):
As I commented elsewhere, when it comes to AI integrations like this, it's probably best to ship and iterate rather than dwindle too long on the first version.

Let's ship this.

@iangmaia have you got a repo in mind where to plug this to stress test?

Also keen to hear if you have any input on this https://github.com/Automattic/dangermattic/pull/111#discussion_r3049243643

lib/dangermattic/plugins/llm_reviewer.rb
```diff
@@ -0,0 +1,385 @@
...
+      end
+
+      findings.empty? ? nil : findings
+    rescue StandardError
+      nil # On any error, fall through to fresh LLM review
+    end
+
+    # Extracts the original tagged message from Danger's inline comment HTML wrapper.
+    def extract_tagged_message(body:)
+      content_cell_match = body.match(%r{<td\b[^>]*data-sticky="(?:true|false)"[^>]*>\s*(.*?)\s*</td>}m)
```
twstokes, line 167: Is this stable? Claude claims that if this starts failing consistently it'll cause extra LLM calls.

The cache flow depends on round-tripping data through Danger's comment HTML:

1. **First run**: `report_single_finding` posts a finding via `warn(msg, file:, line:)`. Danger wraps that message in its inline-comment HTML template — currently something like `<table>...<td data-sticky="false">your message <!-- llm-review:warning --></td>...</table>` — and posts it to GitHub as a PR review comment.

2. **Retry on same SHA**: `load_cached_findings` fetches existing review comments via Octokit, filters to ones whose `commit_id == head_sha` and whose body contains `<!-- llm-review:* -->`, then calls `extract_tagged_message` to recover the original message text from inside the `<td>`:

   ```ruby
   body.match(%r{<td\b[^>]*data-sticky="(?:true|false)"[^>]*>\s*(.*?)\s*</td>}m)
   ```

3. **The hidden dependency**: that regex hardcodes the exact shape of Danger's template — element name `<td>`, attribute name `data-sticky`, attribute value `"true"` or `"false"`. If Danger's `CommentsHelper#generate_inline_comment_body` changes (e.g., switches to `<div>`, renames the attribute, drops quotes, adds a wrapper element, or just rearranges whitespace in a way the `.*?` doesn't tolerate), the regex returns `nil`.

4. **What breaks, silently**: `extract_tagged_message` returning `nil` causes `next unless tagged_message` to skip that comment. Every cached comment gets skipped, `findings` ends up empty, `load_cached_findings` returns `nil`, and `review` falls through to call the LLM again. There's no exception, no warning — the cache just stops hitting.

5. **Why "silently"**:
   - The `rescue StandardError` in `load_cached_findings` would catch a thrown error, but the regex doesn't throw — it just doesn't match.
   - From the user's perspective, Danger still posts the same comments (Danger's own sticky-comment dedup may even hide that they're being re-posted).
   - The only visible signal is your LLM bill — every retry on the same SHA pays for a fresh review, instead of the intended zero-cost replay.

6. **Why it's a real risk**: Danger is an external dep on a separate release cadence. A point release that tweaks the inline-comment template (which is a presentation detail, not part of any public API contract) would break this with no warning. You'd find out from cost monitoring, not from CI failures.

**Mitigations worth considering:**
- Store the canonical message + metadata in a hidden HTML comment you control, e.g., `<!-- llm-review:warning data='base64-encoded-json' -->`, and parse from that instead of from Danger's template. Then you're only coupled to your own format.
- Or, post a separate non-inline marker comment per finding that you fully control, and use inline comments purely for display.
- At minimum, add a unit test that round-trips through the *real* `CommentsHelper.generate_inline_comment_body` (the spec at line 1553 already does this) and pin the Danger gem version, so a bump forces you to re-verify.

twstokes (approved):

lib/dangermattic/plugins/llm_reviewer.rb
iangmaia: Yep this is a tricky point, that's why I was reluctant to merge (plus there was a PR for Danger itself I was waiting to get merged as well, but luckily it already did). I'm hoping to revisit it soon!

PR #112, opened 2026-03-18
Add translation context plugin

iangmaia:
Fixes AINFRA-2164

## Summary

Adds `translation_context_checker`, a Danger plugin that uses the `i18n-context-generator` [gem](https://github.com/Automattic/i18n-context-generator) to generate translator-facing context for changed iOS and Android localization entries.

- **Discovery and context:** `:auto` runs one workflow, preferring changed translation files and otherwise changed source. The plugin passes Danger's base/head refs to the extractor, retains all relevant source evidence, supports free-form context files, and includes the PR title and description as untrusted context by default.
- **Review output:** Results are posted only at PR-changed lines using side-aware locations and safe head-side fallbacks. Fully deleted entries fall back to PR-level feedback, unsupported `.xcstrings` suggestions become normal comments, and multiline source occurrences each receive their own result.
- **Robustness:** Complex and duplicate keys, Android collection members, and comment-only Swift changes are handled deterministically. The plugin validates paths before extraction, runs quietly in CI, and reports configuration, run-level, and per-key failures without hiding successful results or duplicating global failures.

## Test PRs

- https://github.com/woocommerce/woocommerce-android/pull/16302
- https://github.com/woocommerce/woocommerce-ios/pull/17612
- https://github.com/bloom/DayOne-Apple/pull/29307 (.xcstrings catalog)


README.md
```diff
@@ -8,6 +8,21 @@ Add to your project's `Gemfile`
 gem 'danger-dangermattic', git: 'https://github.com/Automattic/dangermattic'
 ```
 
+### Translation context plugin setup
+
+`translation_context_checker` also requires the unpublished `txcontext` gem:
+
+```ruby
+gem 'txcontext', git: 'https://github.com/iangmaia/txcontext'
+```
```
AliSoftware: 🎗️  Reminder to update this once you will have migrated the repo to the Automattic org
(As I can see this PR being merged before the gem repo gets transferred and then forgetting to do a follow-up to update the docs later)

lib/dangermattic/plugins/translation_context_checker.rb
```diff
@@ -0,0 +1,633 @@
...
+    SWIFT_COMMENT_ARGUMENT_PATTERN = /comment:\s*"((?:\\.|[^"\\])*)"/
+
+    # Analyze new or modified translation keys in the PR and suggest context descriptions
+    # to help translators understand how each string is used in the app.
+    #
+    # @param translations [String, Array<String>] Path(s) to translation file(s) (e.g., Localizable.strings, strings.xml).
+    # @param source_paths [String, Array<String>] Path(s) to source code directories to search for string usage.
+    # @param report_location [Symbol, String] (optional) Where to post suggestions. Values: :inline (default), :summary, :both.
+    # @param inline [Boolean, nil] (optional) Deprecated compatibility flag. When provided, overrides report_location together with summary.
+    # @param summary [Boolean, nil] (optional) Deprecated compatibility flag. When provided, overrides report_location together with inline.
```
AliSoftware: Given this has not been published or started to be used officially by any project yet, I don't think we need to keep "compatibility flags" in this API signature?

lib/dangermattic/plugins/translation_context_checker.rb
```diff
@@ -0,0 +1,633 @@
...
+    # @param report_type [Symbol] (optional) Severity for PR-level fallback comments (:message, :warning, :error). Default is :message.
+    #   Only applies when inline placement fails and the comment falls back to a PR-level report.
+    # @param provider [Symbol, String] (optional) LLM provider to use. Default is :anthropic.
+    # @param model [String, nil] (optional) Model name to use. Uses txcontext defaults when omitted.
+    # @param inline_suggestions [Boolean] (optional) Include GitHub suggestion blocks in inline comments when possible. Default is false.
+    # @param inline_suggestion_target [Symbol, String] (optional) Target for inline suggestions. Values: :translation (default), :source.
```
AliSoftware: Those parameters seems to create potential conflicting setups. For example, `report_location: :summary` would make the values of parameters `inline_suggestions` and `inline_suggestions_target` irrelevant. And the value of parameter `inline_suggestion_target` is irrelevant if `inline_suggestions` is `false` to begin with.

💭  What about having a `inline_mode:` parameter with values `:translation_suggestion`, `:source_suggestion`, `:translation_comment`, `:source_comment`, and a `summary:` boolean parameter ?
 - If only `inline_mode` is set to a value while `summary` is false, it would generate inline comment only (either on the source or the translation, either as a plain comment or as a GitHub suggestion )
 - If `inline_mode` is set to `nil` or `:none` or `false`, but we set `summary: true`, only the summary comment will be posted
 - If both are set to a value, both inline comments and summary will be posted (like your `:both` case in current implementation)

iangmaia: Great idea, thanks!! Yeah while adding the different possibilities in iterations, the parameters started to pile up and conflict. Much cleaner: 4695a4f55bd6c4a1aff7f69f243316e8cee80325

lib/dangermattic/plugins/translation_context_checker.rb
iangmaia: 👍 Also updated on https://github.com/Automattic/dangermattic/commit/4695a4f55bd6c4a1aff7f69f243316e8cee80325

lib/dangermattic/plugins/translation_context_checker.rb
AliSoftware: Nice, thanks!

_Sometimes I wish Ruby had "enums with associated values" like Swift does… would allow us to make this API way nicer with `inlineMode: .none` / `inlineMode: .comment(target: .source)` / `inlineMode: .suggestion(target: .translation)` etc 😛  Oh well… 🙃_

lib/dangermattic/plugins/translation_context_checker.rb
```diff
@@ -0,0 +1,621 @@
...
+      none
+    ].freeze
+    STRINGS_KEY_PATTERN = /^\+\s*"([^"]+)"\s*=/
+    XML_STRING_KEY_PATTERN = /^\+.*<string\s+[^>]*?name=["']([^"']+)["']/
+    XML_STRING_ARRAY_KEY_PATTERN = /^\+.*<string-array\s+[^>]*?name=["']([^"']+)["']/
+    XML_PLURALS_KEY_PATTERN = /^\+.*<plurals\s+[^>]*?name=["']([^"']+)["']/
+    XML_FILE_STRING_PATTERN = /<string\s+[^>]*?name=["']([^"']+)["']/
+    XML_FILE_STRING_ARRAY_PATTERN = /<string-array\s+[^>]*?name=["']([^"']+)["']/
+    XML_FILE_PLURALS_PATTERN = /<plurals\s+[^>]*?name=["']([^"']+)["']/
+    SWIFT_COMMENT_ARGUMENT_PATTERN = /comment:\s*"((?:\\.|[^"\\])*)"/
```
AliSoftware, line 88: On one hand I am a RegEx nerd, but on the other hand I feel like parsing via RegEx is a bit fragile… that being said, perfect is the enemy of good, so this is probably a good place to start with at least.

But that made me wonder if we couldn't use more solid solutions to detect strings in the codebase and their changes, especially with native syntax tree parsers API. 🤔 

In particular, if `txcontext` were written in Swift, we could imagine have `txcontext` implement the detection of `NSLocalizedString` calls using `SwiftSyntax` for us, which would be more reliable than RegExes, especially for covering cases of multiline strings (`"""`) or raw strings (`#""`) in Swift's modern syntax that a RegEx could easily miss.

_See my demo project in https://github.com/AliSoftware/swift-syntax-linter-demo and how simple it ends up being to implement detection of `NSLocalizedString` calls in the source code and their parameters using with `SyntaxVisitor` if that helps._

So the idea could be to:
 - Have this Ruby code and Danger plugin detect which source code files have been modified
 - Then pass that list of Swift files modified in the PR to a subcommand of `txcontext`—so that it would run SwiftSyntax only on those `.swift` files.
 - That subcommand would then output e.g. a JSON listing each detected call sites of `NSLocalizedString` that appeared in that `.swift` file, with for each the line and column numbers and the extracted key, value and comment
 - Then the Ruby code would use the line/column numbers information from that JSON to compare that with the PR diffs to only focus on the call sites that overlap with that diff and thus correspond to `NSLocalizedStrings` calls having changed in the PR.
 - And ultimately call `txcontext` on those like you're doing now when you're using RegEx to do that filtering/detection

Or instead of having a back-and-forth between `txcontext` listing the detected calls, then Ruby filtering based on diff, then `txcontext` being called again with the sublist… we could have everything implemented in the main `txcontext` tool, where we'd pass the diff (or at least the list of modified files + the list of line ranges that appear modified/added in the diff) as additional parameter in a single call…

---

Of course all that suppose that `txcontext` would be implemented in Swift, while currently in practice it's in Ruby. But:
1. Either this could be an evolution idea for a v2 of the tool (to migrate its implementation from Ruby to Swift so it can use SwiftSyntax) to improve the reliability of the strings detection and support multiline & raw strings etc in a v2
2. Or maybe you could keep `txcontext` in Ruby but implement the call to `SwiftSyntax.SyntaxVisitor` (like I did in my demo project) in a native extension to the gem?
3. Or maybe there's a way to call Swift Syntax or some `swift --parse-ast` command or whatever from the command line—thus **not** requiring to have a separate binary implemented in Swift so it can use SwiftSyntax, but instead just use an existing command line tool. For example if we `brew install sourcekitten`, a call like `sourcekitten structure --file file.swift` would already be able to output the Swift code AST structure as a JSON that your Ruby code can then use to detect the `NSLocalizedString` call sites…
4. Or rely on a source code parser like [TreeSitter](https://github.com/tree-sitter/tree-sitter) or similar (the kind of tools that is used to implement autocomplete in IDEs on top of it etc) and [its Ruby bindings](https://github.com/Faveod/ruby-tree-sitter?tab=readme-ov-file) to parse the Swift source code (using [this swift grammar](https://github.com/alex-pinkus/tree-sitter-swift)) and have a tree representation that would make the parsing and call sites detection more reliable while not requiring Swift binaries (but likely requiring `libffi` bindings to the C libs

Obviously, except if you're ready to re-implement `txcontext` in Swift rather than Ruby (which might also bring the question about cross-platform portability), this is probably very much ideas for a v2 or even v3. But I thought this could be interesting brain dump and ideas to share, just to open the debate about better supporting edge cases (like multiline-strings and raw-strings).

lib/dangermattic/plugins/translation_context_checker.rb
```diff
@@ -0,0 +1,621 @@
...
+        return
+      end
+
+      translations = Array(translations)
+      source_paths = Array(source_paths)
+
+      # Only process translation files that were changed in this PR
+      changed_translation_files = translations.select do |path|
+        git_utils.added_and_modified_files.include?(path)
+      end
```
AliSoftware: Wait, doesn't that mean that this plugin will only trigger on changes to `Localized.strings` files… which would typically only change during code-freeze when we call `genstrings` on the source code to extract the strings into said `.strings` file?

While I'd expect that if I have a PR that only changes a `.swift` file to add or modify a call to `NSLocalizedString(…)` somewhere in the codebase… without the PR making any change to the `Localized.strings` file (which isn't expected in our typical development workflow on iOS repos at a8c), then Danger would not consider it? And only when we go to the code freeze and generate the `Localizable.strings` file will Danger finally detect the new or modified entries in the `.strings` file, try to find the corresponding string in the source code, and do its thing form there?

lib/dangermattic/plugins/translation_context_checker.rb
AliSoftware: Discussed this with Claude who had an interesting take: https://claude.ai/share/191757f4-810a-4eaf-ab78-6d9ce0237729

iangmaia: >  this is probably very much ideas for a v2 or even v3. But I thought this could be interesting brain dump and ideas to share

Yep, thanks for sharing!  While running tests, some issues came up and a more sophisticated mechanism crossed my mind, but indeed this was more to validate the idea and to have something that would work in most cases.

> perfect is the enemy of good, so this is probably a good place to start with at least

💯 

> Claude who had an interesting take

Indeed! Definitely worth looking into.

lib/dangermattic/plugins/translation_context_checker.rb
iangmaia: This is still required for Android, where the context goes to the translations.
Though I've implemented some tweaks on the "source" mode, which is targeted for iOS: https://github.com/Automattic/dangermattic/pull/112/commits/514c3418c0ef3c940321098ec1e7562aa5d77442.

lib/dangermattic/plugins/common/inline_markdown_poster.rb
```diff
@@ -0,0 +1,137 @@
...
+    def danger_supports_ranged_inline_markdown?
+      Danger::Markdown.instance_method(:initialize).parameters.any? do |type, name|
+        %i[key keyreq].include?(type) && name == :start_line
+      end
+    end
+
+    # Temporary bridge until Danger can post ranged inline markdown comments.
+    # Callers use `post` either way, so migration back to pure Danger is a
+    # helper-internal change once upstream support lands.
+    def upsert_raw_github_review_comment!(markdown, file:, line:, start_line:, side:, start_side:)
```
iangmaia, line 72: FYI: this part is what makes me reluctant to already put this PR as ready for review and merge it, as I consider it temporary.

Ideally, the PR https://github.com/danger/danger/pull/1542 should be merged and most (if not all) of this `inline_markdown_poster` plugin can be removed.

iangmaia: The Danger PR has landed and is available from Danger's release https://github.com/danger/danger/releases/tag/v9.6.0

mokagio:
@iangmaia , my code review scout agent noticed a couple of possible issues. https://github.com/Automattic/dangermattic/pull/144 demonstrates them.

iangmaia:
@mokagio hopefully addressed the two regressions demonstrated in #144:

- Preserved `.strings` and `strings.xml` entries when replacing a same-line trailing translator comment
- Kept the added-line numbering correct when hunk content begins with `++`

I cherry-picked #144’s two regression-test commits then added the fixes in `119ab2b`.

iangmaia:
Final generator integration is now pushed in 9f7dc05:

- require i18n-context-generator ~> 0.5.2 so existing 0.5.1 locks cannot retain the buggy parser
- resolve Gemfile.lock to the published 0.5.2 gem
- update the setup documentation minimum version

Verification:
- bundle exec rake
- 380 examples, 0 failures
- 46 RuboCop-clean files
- Danger plugin lint passed

Generator release: https://github.com/Automattic/i18n-context-generator/releases/tag/0.5.2

mokagio (approved):
Good on my end. Thanks @iangmaia.

@AliSoftware might want to chip in to follow up on his comments that have been addressed.

README.md
```diff
@@ -8,6 +8,52 @@ Add to your project's `Gemfile`
...
+
+Dangermattic requires `i18n-context-generator` 0.5 or newer. Bundler installs the released gem automatically
+with Dangermattic.
+
+Expose `ANTHROPIC_API_KEY` in CI so `i18n-context-generator` can generate context suggestions:
+
+```yaml
+env:
+  ANTHROPIC_API_KEY: "${ANTHROPIC_API_KEY}"
+```
```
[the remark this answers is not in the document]
AliSoftware: Probably a good point from Copilot here?

Also, we should double-check that this API KEY won't leak in the logs when provided that way.

I _think_ it should be obfuscated automatically by [Buildkite's own secret redactor](https://buildkite.com/docs/agent/cli/reference/redactor) and the default `--redacted-vars` list including the `*_API_KEY` pattern. But still worth double-checking with a practical test.

README.md
```diff
@@ -8,6 +8,52 @@ Add to your project's `Gemfile`
 gem 'danger-dangermattic', git: 'https://github.com/Automattic/dangermattic'
 ```
 
+### Translation context plugin setup
+
+Dangermattic requires `i18n-context-generator` 0.5.2 or newer. Bundler installs the released gem automatically
+with Dangermattic.
```
AliSoftware: This wording is confusing to me, first because I wasn't sure what this `i18n-context-generator` was (is it a `gem`? A binary tool?), and also the "installs the _released_ gem" made me confused as to why it has to specifically say "released" there.


```suggestion
Dangermattic depends on the [`i18n-context-generator`](https://github.com/Automattic/i18n-context-generator) gem version `0.5.2` or newer. Bundler installs it automatically as a dependency of Dangermattic.
```

README.md
```diff
@@ -8,6 +8,52 @@ Add to your project's `Gemfile`
...
+
+Expose `ANTHROPIC_API_KEY` in CI so `i18n-context-generator` can generate context suggestions:
```
AliSoftware:
```suggestion
First, you need to expose `ANTHROPIC_API_KEY` in your CI pipeline, so `i18n-context-generator` can generate context suggestions:
```

README.md
```diff
@@ -8,6 +8,52 @@ Add to your project's `Gemfile`
...
+
+```yaml
+env:
+  ANTHROPIC_API_KEY: "${ANTHROPIC_API_KEY}"
+```
+
+For `translation_context_checker`, use `discovery_mode: :source` for code-first flows like iOS and
+`discovery_mode: :translations` for resource-first flows like Android. `source_paths` should always be set
```
AliSoftware: IMHO this is not super clear. What does `code-first flow`and `resource-first` flow really means in this context?

After thinking a bit about it, I think I understand what you might mean and try to convey here, but I still think this requires a better explanation of what those 2 different discovery_mode are.


```suggestion
For `translation_context_checker`, use `discovery_mode: :source` for code-first flows like iOS where the source code is the source of English copies, and
`discovery_mode: :translations` for resource-first flows like Android where `values/strings.xml` is the source of English copies. `source_paths` should always be set
```
And in fact, re-writing it like this makes me wonder, does `discovery_mode: :translations` suggests that it will try to discover strings in **translations** only, namely only in `values-[locale]/strings.xml` for Android… but not in **source English copies** like `values/strings.xml`? Or is `:translations` here a red-herring and potentially poorly chosen name for the constant, where `:resource_files` would be a better description to suggest "we'll search English copy strings in resource files not source code"?

Or does `discovery_mode:` is not just for discovering English copies (e.g. `values/strings.xml`), but _also_ scans translations (`values-[locale]/strings.xml`) too and use those translations for other purposes (to refine the context based on what might already be translated)? And if that's the case, shouldn't it be doing the same for iOS where it discovers English copies in the source code… but translations still live in `.strings` resource files? Though I doubt `discovery_mode` goes that far in practice, so maybe that last paragraph is not relevant.

README.md
```diff
@@ -8,6 +8,52 @@ Add to your project's `Gemfile`
...
+
+The plugin includes the pull request title and description as untrusted model evidence by default. Set
+`include_pull_request_context: false` to disable that behavior. The default `:auto` mode performs exactly one
+extraction: translation-backed discovery takes priority when a configured translation file changed; otherwise
+it uses source-backed discovery when a configured source file changed. Explicit modes run only when their
+corresponding files changed. Extraction failures are reported as one aggregate warning while successful
+suggestions are still shown.
+
+In a mixed PR, `:auto` intentionally does not run a second source-backed pass, so localization calls that exist
+only in changed source are not included. Run the plugin twice with explicit modes when both workflows are wanted:
```
AliSoftware: TBH this gave me a headache.

I think we can do better at explaining how the whole plugin (and especially how each of its parameters) work, with concrete examples first and bullet point explaining each parameter's role one by one, rather than those intricate and hard-to-decipher long, obviously AI-generated paragraphs.

README.md
```diff
@@ -8,6 +8,52 @@ Add to your project's `Gemfile`
...
+  source_paths: ['Sources/'],
+  context_files: ['GLOSSARY.md', 'docs/localization-style.md']
+)
+```
+
+The extractor uses the base and head refs prepared by Danger, so it shares Danger's merge-base behavior in
+shallow CI clones. Relevant source snippets are sent to the configured external LLM provider. Do not enable
+this check for source that your provider is not permitted to process. Context files and pull request metadata
+are sent under the same redaction and prompt-injection-resistant evidence boundary; context files are included
+in full and must fit the configured prompt limit.
```
AliSoftware: Ditto, reading this gives me a headache trying to even understand what this whole paragraph is trying to say.

CHANGELOG.md
```diff
@@ -10,15 +10,34 @@ _None_
...
+- Use machine-readable extractor result states instead of matching placeholder descriptions.
+- Group changed source locations by localization occurrence so multiline calls produce one inline result.
+- Preserve left/right diff-side metadata, using head fallbacks for apply-ready
+  suggestions when translator comments are removed.
+- Fall back to plain inline feedback for `.xcstrings` and other formats that do
+  not support one-click comment suggestions.
+- Include translation-file identity in summary rows and publish inline feedback
+  in deterministic file/key/line order.
+- Preserve translation entries when replacing same-line translator comments and
+  keep added-line numbering correct for header-like diff content.
```
AliSoftware: Given this whole `translation_context_checker` plugin is brand new, what's the point of listing all those under Bug Fixes? I'm guessing those are bugs you fixed as part of working and iterating on this PR, but not of bugs that ever made it into `dangermattic`'s latest release `1.4.1`, so nothing that was a bug in `1.4.1` that this new release is fixing.



```suggestion
```

lib/dangermattic/plugins/translation_context_checker.rb
```diff
@@ -0,0 +1,332 @@
...
+  # Suggests translator context for localization entries changed in a pull request.
+  #
+  # Relevant source snippets are sent to the configured external LLM provider.
+  # The matching provider API key must be available in CI.
+  #
+  # @example Translation-backed suggestions
+  #   translation_context_checker.check_context_suggestions(
+  #     discovery_mode: :translations,
+  #     source_paths: ['Sources/'],
+  #     translation_paths: ['Resources/Localizable.strings'],
```
AliSoftware: So, in the `README.md` you say that typically for iOS, which uses source code flow and provides the source of truth of English copy via source code, you're supposed to use `discovery_mode: :source`.

But then this first example you show an example of Translation-backed suggestions…
Which is confusing (when would it be useful to use this for an iOS project?) and also highlight my other remark that it seems that this does not use the "translations" (as in "translations of English copies in another language") but rather the "resource file" (as in "the `Localizable.strings` resource file in `Base.lproj` or `en.lproj` containing the English copies"), so maybe `discovery_mode: :translations` is a misleading/confusing mode name.

README.md
```diff
@@ -8,6 +8,52 @@ Add to your project's `Gemfile`
...
+
+```yaml
+env:
+  ANTHROPIC_API_KEY: "${ANTHROPIC_API_KEY}"
+```
+
+For `translation_context_checker`, use `discovery_mode: :source` for code-first flows like iOS and
+`discovery_mode: :translations` for resource-first flows like Android. `source_paths` should always be set
+explicitly because the generator still searches source code for usage context, including when
+`discovery_mode` is `:translations` or `:auto`. `translation_paths` is only for translation-backed runs.
```
AliSoftware: If `translation_paths:` is only for translation-backed runs, then do we really need to specify a `discovery_mode:` parameter after all?

Couldn't we consider that if `translations_paths:` is provided and set to a value, then it's gonna be discovering the English copies from that `translations_paths:` (e.g. the Android case), and if it's not set or `nil` then it's gonna discover the English copies from the source code (e.g. the iOS case)?

Speaking of which, if `translations_path:` is supposed to point to the `values/strings.xml` resource file, which contains the English copies, to me naming this "translations path" is misleading, because to me "translations" suggestions "translations of the English copies into another locale".
Also, why is this attribute plural? Does that mean that we're expected to provide the paths to resources for **all** translations in the various locales, e.g. `['src/main/res/values/strings.xml'] + Dir.glob('src/main/res/values-*/strings.xml')`?

If we're only expected to pass the English string resource, "English string resource path" or `source_strings_resource_path:` would be a better description of this, even if a bit of a mouthful. Or anything like `strings_discovery_resource_path:` or something. Then iOS would just not provide it (or provide an explicit `nil`, or maybe we can support a `:source_code` magic value if you feel it'd be more explicit) to indicate it would scan the source code instead; and we don't need a `discovery_mode:` parameter anymore, avoiding conflicting configurations to ever be made possible.

lib/dangermattic/plugins/translation_context_checker/location_resolver.rb
```diff
@@ -0,0 +1,348 @@
...
+        lines = cached_file_lines(location[:file])
+        next location.merge(content: '', inline_target: :translation) if location[:side] == 'LEFT'
+        next unless lines && location[:line].between?(1, lines.length)
+
+        location.merge(content: lines[location[:line] - 1], inline_target: :translation)
+      end
+
+      locations.uniq do |location|
+        [location[:file], location.fetch(:original_side, location[:side])]
+      end
```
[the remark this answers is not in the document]
AliSoftware: I think this is a good point worth fixing?

CHANGELOG.md
```diff
@@ -10,15 +10,34 @@ _None_
...
 
 ### Internal Changes
 
-_None_
+- Require Danger 9.6 and use its native ranged Markdown support instead of a custom GitHub posting layer.
+- Run at most one extractor workflow per plugin invocation and delegate translation diff parsing to the extractor.
+- Read file diffs only when constructing apply-ready inline suggestions.
+- Run the extractor quietly and split extraction, location resolution, suggestion
+  rendering, and publication into focused components.
```
AliSoftware: Ditto, those seems to be changes  you made while iterating on this PR, not changes made to things that already existed in Dangermattic version `1.4.1`?

README.md
AliSoftware: And in case I'm misunderstanding the set of possible option combinations that we might pass (and if that's the case, that's also a sign that the README / documentation is confusing / not clear enough), but instead that some of those combinations would still make sense (e.g both passing `discovery_mode: :source` **and** `translations_paths: Dir.glob(File.join('Resources', '*.lproj', 'Localizable.strings'))` list of all English + translation resource files? 😵 ), then maybe using subclasses as configuration pattern would make more sense (a bit like how we did it in `release-toolkit` with the `VersionCalculator` and `VersionFormatter` subclasses, if that makes sense)?

For example `discovery_mode: SourceDiscoveryMode.new` vs `discovery_mode: ResourceFileDiscoveryMode.new(path: …)` etc. The drawback of this is that the parameters to the Danger DSL for this plugin would not be only basic Ruby types (like Int, String, Hash, …) but rely on custom types, but if that makes the API less confusing and reduce the complexity of the API surface and avoid the risk of conflicting parameters, maybe worth considering? I'm ok if we could keep  more standard Ruby types instead (e.g. Symbol and Hash), as long as we manage to get an API surface whose complexity is acceptable and not confusing nor risking to pass conflicting parameters too easily (especially if each parameter's role and interdependence is not clear enough from the doc to begin with)

AliSoftware (commented):
I have only re-reviewed the docs (README, CHANGELOG, doc comments) so far, and I already find a lot of it confusing, with some parts making me wonder if the API of the plugin and of the `I18nContextGenerator::Config` were not more complex that they needed to be (misleading config option names, inter-dependent config options that could end up conflicting while maybe they could be exposed with a single non-conflicting parameter or some other shape…)

I haven't yet reviewed the ruby code itself yet (nor the specs) given the feedback I've left on those above points might change the shape of the code anyway.

iangmaia:
@AliSoftware @mokagio 
> I already find a lot of it confusing, with some parts making me wonder if the API of the plugin and of the `I18nContextGenerator::Config` were not more complex that they needed to be
> 
> I haven't yet reviewed the ruby code itself yet (nor the specs) given the feedback I've left on those above points might change the shape of the code anyway.

I agree -- this started as a small experiment that was iterated on and off (sometimes with a couple of months in between) and it could use some updates. I've pushed some improvements today, but I'll also give it a better look in the coming days.

PR #126, opened 2026-06-15
Add `check_existing_strings_not_modified` to `android_strings_checker`

oguzkocer:
New check that fails when an **existing** translatable `<string>`'s value is changed in place instead of added under a new key. This enforces the "new copy → new key" convention so in-progress translations stay valid as we move to continuous localization.

- Diff-based: a key in both the removed and added lines with a different value is a violation. Add / remove / rename / reorder are allowed; `translatable="false"` is ignored.
- Default selector targets source `…/res/values/strings.xml` only (skips localized `values-<locale>/` and generated copies). `report_type:` defaults to `:error`.

---

I tried to test this PR in https://github.com/wordpress-mobile/WordPress-Android/pull/22986, but it doesn't look like I can point to this branch, so I think it needs to be reviewed and landed first. Let me know if there is a testing process that I'm missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-- Added `android_strings_checker.check_existing_strings_not_modified`, which fails when the value of an existing translatable `<string>` is changed in place (rather than added under a new key). This enforces string-key immutability, which keeps in-progress translations valid in a continuous-localization setup.
+- Added `android_strings_checker.check_existing_strings_not_modified`, which fails when the value of an existing translatable `<string>` is changed in place (rather than added under a new key). This enforces string-key immutability, which keeps in-progress translations valid in a continuous-localization setup. [#126]
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


mokagio (approved):
Tested locally by:

1. Checking out https://github.com/wordpress-mobile/WordPress-Android/pull/22986
2. Running `bundle install && bundle exec danger pr https://github.com/wordpress-mobile/WordPress-Android/pull/22986` 

Result:

Image

I'm not sure who the formatting will render in the GitHub comment, but that's something we can tweak afterwards if necessary.

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-_None_
+- Added `android_strings_checker.check_existing_strings_not_modified`, which fails when the value of an existing translatable `<string>` is changed in place (rather than added under a new key). This enforces string-key immutability, which keeps in-progress translations valid in a continuous-localization setup.
```
mokagio:
```suggestion
- Added `android_strings_checker.check_existing_strings_not_modified`, which fails when the value of an existing translatable `<string>` is changed in place (rather than added under a new key). This enforces string-key immutability, which keeps in-progress translations valid in a continuous-localization setup. [#126]
```

PR #128, opened 2026-06-16
Release 1.3.0 into trunk

oguzkocer:
Releasing new version 1.3.0.

# What's Next

PR Author: Be sure to create and publish a GitHub Release pointing to `trunk` once this PR gets merged,
copy/pasting the following text as the GitHub Release's description:
```
### New Features

- Added `android_strings_checker.check_existing_strings_not_modified`, which fails when the value of an existing translatable `<string>` is changed in place (rather than added under a new key). This enforces string-key immutability, which keeps in-progress translations valid in a continuous-localization setup. [#126]


```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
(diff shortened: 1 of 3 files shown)


mokagio (approved):
Thanks for following up on this @oguzkocer !

I would have liked to do it yesterday after merging your PR, but got sidetracked by other work.

PR #130, opened 2026-06-24
Add bundler block to Dependabot config and route Ruby reviews

mokagio:
Updates Dependabot to check Ruby dependencies daily so we can more easily address security updates. See [AINFRA-2437](https://linear.app/a8c/issue/AINFRA-2437).

Also routes review of the Apps Infra surface to @Automattic/apps-infra-tooling via CODEOWNERS. GitHub retired the dependabot.yml `reviewers` key, and CODEOWNERS is its designated replacement.

---

Posted by Codex (GPT-5) on behalf of @mokagio with approval.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,13 @@
+# Ruby dependency surface
+Gemfile*              @Automattic/apps-infra-tooling
+*.gemspec             @Automattic/apps-infra-tooling
+.bundle/              @Automattic/apps-infra-tooling
+.rubocop*.yml         @Automattic/apps-infra-tooling
+
+# CI and automation
+.buildkite/           @Automattic/apps-infra-tooling
+.github/workflows/    @Automattic/apps-infra-tooling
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,13 @@
...
+.bundle/              @Automattic/apps-infra-tooling
+.rubocop*.yml         @Automattic/apps-infra-tooling
+
+# CI and automation
+.buildkite/           @Automattic/apps-infra-tooling
+.github/workflows/    @Automattic/apps-infra-tooling
+.github/dependabot.yml @Automattic/apps-infra-tooling
+
+# Toolchain and environment pins
+.ruby-version         @Automattic/apps-infra-tooling
```
mokagio, line 13: My automation used the same heuristic as for projects that we support. But given this is a project we own entirely, I wonder whether we should have a stricter configuration. No sure if it's supported but something like: `* @Automattic/apps-infra-tooling`

For sure, this is a place to start. We can always update later.

iangmaia (approved):

PR #135, opened 2026-07-09
Add reusable Dependabot auto-merge workflow

mokagio:
Woo Android, Day One Android and Studio each grew their own copy of the same workflow: fetch Dependabot's metadata, gate on `update-type`, approve, enable auto-merge. This lifts that into a reusable workflow so the mobile repos with no auto-merge today get it by adding a caller, and the three that have one can drop their local copies.

Patch updates auto-merge; everything else still waits for a human.

Part of [AINFRA-2617](https://linear.app/a8c/issue/AINFRA-2617). Adoption is tracked per-repo under the [project](https://linear.app/a8c/project/implement-auto-merge-for-dependabot-prs-via-reusable-gha-workflow-526a1c6e74f3/overview).

### Intentional tradeoffs

**The `denylist` and `minor-update-allowlist` inputs ship now, though only patch auto-merge is needed.** Both default to `[]`, so a repo configuring nothing gets patch-only behaviour. They're here because the expensive part is computing the decision in one step instead of repeating `if: steps.metadata.outputs.update-type == …` on every step — once that exists, each list is a `jq` lookup. Adding them later would mean doing that restructure twice.

**No checkout step.** Neither `dependabot/fetch-metadata` nor `gh` needs a working tree. The three prior implementations all paid for one.

**`pull_request`, not `pull_request_target`** (Studio uses the latter). Since nothing is checked out the extra privilege buys nothing, and `pull_request` is the documented path. Consequence: Dependabot runs can't read Actions secrets, so an override token must be stored as a *Dependabot* secret. Documented in the README.

**Milestone assignment is deliberately absent**, though Woo Android's implementation has it. It's only needed where a `Dangerfile` reports a missing milestone as an `:error` (Woo iOS and Android), not where it takes the plugin's default `:warning` (WordPress iOS, Pocket Casts iOS). Woo can unblock by having Dependabot apply the `milestone-not-required` label their `Dangerfile` already honours. Assigning milestones for release tracking is a separate concern, explored in #136.

### Gotchas for adopters

- The **caller** must grant `contents: write` and `pull-requests: write` — a reusable workflow can't hold more permission than its caller.
- The repo needs **"Allow auto-merge"** on, and the org must **allow Actions to approve PRs**, or the workflow fails at the approve step. Verified enabled for the four prototype repos.

### How to test

`bundle exec rake` covers the repo's own checks, but not this workflow. I verified the decision logic out-of-band against a 16-case matrix (patch/minor/major, grouped updates, deny beats allow, exact-not-prefix name matching, malformed inputs, a `merge-method` injection attempt); `shellcheck` is clean.

That verification isn't committed here, so it protects nothing. **#137 stacks on this branch and adds it as RSpec**, driving the decision script straight out of the workflow YAML so a renamed input breaks the build. Review order: this PR, then #137.

Real end-to-end validation happens in the prototype repos (dangermattic itself, AutoProxxy, hostmgr, apps-infra-toolbox), which point at this branch before a tag is cut.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -83,4 +83,72 @@ This workflow runs Danger directly on GitHub Actions.
      - Executes Danger in read-only mode for forks and Dependabot PRs
      - Runs Danger with full functionality for PRs where the configured token has access to the repo
 
-These reusable workflows can be incorporated into other workflows in your repository to perform specific tasks related to issue labeling, Buildkite job management, and pull request checks using Danger.
+## Auto-Merge Dependabot Pull Requests
+
+**File:** `workflows/reusable-dependabot-auto-merge.yml`
+
+This workflow approves and enables auto-merge on Dependabot pull requests.
+By default it only does so for patch updates; everything else is left for a human to review.
+
+Auto-merge is enabled rather than merging directly, so the pull request still has to pass the repository's required checks before it lands.
+
+### Inputs:
+- `merge-method`: The merge method to use, one of `merge`, `squash` or `rebase` (default: `merge`)
+- `minor-update-allowlist`: JSON array of dependency names that may also be auto-merged on minor updates (default: `[]`)
+- `denylist`: JSON array of dependency names that are never auto-merged, whatever the update type (default: `[]`)
+- `assign-closest-milestone`: Assign the closest open milestone with a future due date before merging (default: `false`)
+  - Useful where Danger requires a milestone, since the pull request would otherwise never satisfy its required checks.
+
+Example:
+
+```yaml
+with:
+  minor-update-allowlist: |
+    [
+      "release-toolkit"
+    ]
+  denylist: |
+    [
+      "some-fragile-dependency"
+    ]
+```
+
+For grouped updates, a denylisted dependency anywhere in the group blocks the pull request, and a minor update is only auto-merged when every dependency in the group is on the allowlist.
+Names are matched exactly, so `okhttp` on the denylist does not block `okhttp-urlconnection`.
+
+### Secrets:
+- `github-token`: Optional GitHub token, defaulting to `GITHUB_TOKEN`
+  - Dependabot-triggered runs cannot read Actions secrets, so a token passed here must be stored as a [Dependabot secret](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/configuring-access-to-private-registries-for-dependabot).
+
+### Job: `dependabot-auto-merge`
+- Permissions: `contents: write`, `pull-requests: write`
+  - A reusable workflow cannot hold more permissions than its caller, so the calling workflow must grant these too.
+- Runs only for Dependabot pull requests opened from a branch on the repository itself, never from a fork.
+- Steps:
+  1. Fetch Dependabot metadata
+  2. Decide whether to auto-merge, from the update type and the allow/deny lists
+  3. Assign the closest milestone, when enabled
+  4. Approve the pull request
+  5. Enable auto-merge
+
+The calling workflow must use the `pull_request` event and grant the permissions above:
+
+```yaml
+name: 🤖 Auto-merge Dependabot Updates
+
+on:
+  pull_request:
+    types: [opened, reopened]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  dependabot-auto-merge:
+    uses: Automattic/dangermattic/.github/workflows/reusable-dependabot-auto-merge.yml@v1.5.0
+```
+
+The repository must have "Allow auto-merge" enabled, and the organisation must allow GitHub Actions to approve pull requests, otherwise the workflow fails when approving or enabling auto-merge.
+
+These reusable workflows can be incorporated into other workflows in your repository to perform specific tasks related to issue labeling, Buildkite job management, Dependabot updates, and pull request checks using Danger.
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -125,7 +125,7 @@ Names are matched exactly, so `okhttp` on the denylist does not block `okhttp-ur
 - Steps:
   1. Fetch Dependabot metadata
   2. Decide whether to auto-merge, from the update type and the allow/deny lists
-  3. Approve the pull request
+  3. Approve the pull request, leaving a review comment saying why it was auto-approved
   4. Enable auto-merge
 
 The calling workflow must use the `pull_request` event and grant the permissions above:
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-dependabot-auto-merge.yml
```diff
@@ -0,0 +1,164 @@
...
+            echo '⚠️ No open milestone with a future due date. Leaving the pull request unmilestoned.'
+            exit 0
+          fi
+
+          echo "🗓️ Assigning milestone '$milestone'"
+          gh pr edit "$PR_NUMBER" --milestone "$milestone"
+
+      - name: 👍 Approve the Pull Request
+        if: steps.decision.outputs.should-merge == 'true'
+        run: gh pr review --approve "$PR_NUMBER"
```
AliSoftware: Can we make the review leave a comment saying something like "Auto-approved by Danger as this is a non-major dependency update" or something?

.github/workflows/README.md
```diff
@@ -83,4 +83,72 @@ This workflow runs Danger directly on GitHub Actions.
...
 
-These reusable workflows can be incorporated into other workflows in your repository to perform specific tasks related to issue labeling, Buildkite job management, and pull request checks using Danger.
+## Auto-Merge Dependabot Pull Requests
+
+**File:** `workflows/reusable-dependabot-auto-merge.yml`
+
+This workflow approves and enables auto-merge on Dependabot pull requests.
+By default it only does so for patch updates; everything else is left for a human to review.
+
+Auto-merge is enabled rather than merging directly, so the pull request still has to pass the repository's required checks before it lands.
```
AliSoftware, line 93: 👍

.github/workflows/reusable-dependabot-auto-merge.yml
```diff
@@ -0,0 +1,164 @@
...
+on:
+  workflow_call:
+    inputs:
+      merge-method:
+        description: The merge method to use. One of `merge`, `squash` or `rebase`.
+        default: merge
+        type: string
+        required: false
+      minor-update-allowlist:
+        description: Dependencies that may also be auto-merged on minor updates; must be a JSON array of strings.
```
mokagio, line 12:
```suggestion
        description: Dependencies that may also be auto-merged on minor updates. Must be a JSON array of strings.
```

.github/workflows/reusable-dependabot-auto-merge.yml
```diff
@@ -0,0 +1,164 @@
...
+        default: |
+          []
+        type: string
+        required: false
+      denylist:
+        description: Dependencies that must never be auto-merged, whatever the update type; must be a JSON array of strings.
```
mokagio, line 18:
```suggestion
        description: Dependencies that must never be auto-merged, whatever the update type. Must be a JSON array of strings.
```

AliSoftware (commented):
Feels a little strange to have such reusable GitHub Action workflow be hosted in the dangermattic repo while it's not doing anything with Danger or dangermattic?

OTOH not sure which place would be more appropriate for it to live.

But that made me wonder if this couldn't be a Dangermattic plugin instead, making it a command to call inside your `Dangerfile` like `dependabot.automerge_trivial_updates(minor_allow_list: …, …)` instead of being a GitHub Action?

mokagio:
Thanks for the review @AliSoftware .

The AI should not have opened this as ready for review, as I'm still iterating with it on the implementation details. At the same time, it's great to have early feedback so I can fold it in.

---

> Feels a little strange to have such reusable GitHub Action workflow be hosted in the dangermattic repo while it's not doing anything with Danger or dangermattic?

That's a fair point. I went with it because we already have some shared workflows in here, e.g. https://github.com/Automattic/dangermattic/blob/0fe996e93ce06c2dabcc776ea45fd04ff0ffe670/.github/workflows/reusable-check-labels-on-issues.yml which does not run Danger (however, it's Danger-like in its nature).

It's fine for me to move to a dedicated repo, something like `apps-infra-gha-workflows`.

.github/workflows/reusable-dependabot-auto-merge.yml
mokagio: Good call — done in 2536b2f.

The decision step already worked out *why* it was letting a PR through (it logs it), so I threaded that through as a step output rather than hardcoding a string. The approval now reads:

> 🤖 Auto-approved by Dangermattic's Dependabot auto-merge workflow, because all updates are patch level.

or, for an allowlisted minor:

> 🤖 Auto-approved by Dangermattic's Dependabot auto-merge workflow, because every dependency in this minor update is on the allowlist.

That matters more once repos configure the allow/deny lists differently — a bare bot approval doesn't tell a reader which rule applied, and the rules won't be the same everywhere.

One deviation from your wording: I avoided saying "by Danger", since this workflow doesn't run Danger. It's a plain GitHub Actions workflow that happens to live in Dangermattic.

Covered by tests in #137, which asserts both the wording the decision step emits and that the approval step reads it into `--body` — dropping either side fails the build.

---

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

AliSoftware:
I'm not sure it warrants a new dedicated repo, especially since the existing ones in this repo all altar either relate to dangermattic, or like the one you link to, at least relate to _PR/issue checks_, and thus also to something similar enough to what the spirit of what danger does. Do this new one for Dependabot would be the only outlier, and creating a repo just for hosting one workflow is probably not worth it at this stage 😅 unless we envision/apart know there will be more workflows added to it soon.

But that is also why I suggested to also experiment with implementing this "Dependabot automerge" feature as a Dangermattic plugin rather than a GitHub workflow so it could fit better here, I think it's worth a try?

mokagio:
> But that is also why I suggested to also experiment with implementing this "Dependabot automerge" feature as a Dangermattic plugin rather than a GitHub workflow so it could fit better here, I think it's worth a try?

Worth a try for sure, and cheap to get AI to stub it out.

I have a couple of reservations, neither particularly strong, mind you. It will be good to see how the Danger-native way turns out and whether it makes the trade worth.

- Danger is "unit tests for pull requests." Making it take actions such as approving and auto-merging seems out of place. The fact that it should be possible because of the direct GitHub API access doesn't mean it's appropriate.
- A GHA can run on any repo, without concerns for setting up Ruby gems or Danger. This makes it a bit easier to adopt, especially for the wider org. One workflow vs Ruby stack + Danger + Buildkite (+ retrigger workflow maybe?)

Let's see how things turn out!

(One advantage of Danger/Ruby-native is testability. But, in talking with Opus 4.8, I discovered a possible way to test the YAML workflows which is being explored in https://github.com/Automattic/dangermattic/pull/137. I still think that with Danger it would be more robust, but at least it's not a tests/no-tests situation)

iangmaia:
> I'm not sure it warrants a new dedicated repo
> [...]
> Do this new one for Dependabot would be the only outlier, and creating a repo just for hosting one workflow is probably not worth it at this stage 😅 

There's also a potential shared workflow for the Claude Code review caller.

FYI the Woo org created [this](https://github.com/woocommerce/.github/tree/trunk/.github/workflows) (so another repo in the end). I think we could start either way.

> But that is also why I suggested to also experiment with implementing this "Dependabot automerge" feature as a Dangermattic plugin rather than a GitHub workflow so it could fit better here, I think it's worth a try?

Though the code looks ok, I agree with Gio (posted a comment [here](https://github.com/Automattic/dangermattic/pull/138#issuecomment-4967363810)) it feels a bit off.

PR #137, opened 2026-07-09
Test the Dependabot auto-merge decision logic

mokagio:
> **Stacked on #135.** Base is `mokagio/ainfra-2617-dependabot-auto-merge`, so this diff shows only the tests. Merge #135 first and GitHub will retarget this to `trunk`.

#135 ships the auto-merge decision as inline bash, matching `reusable-check-labels-on-issues.yml`. I verified it out-of-band against a case matrix, but that verification wasn't committed, so nothing stopped a future edit from silently inverting the allow/deny semantics. This commits it.

### Why it reads the YAML instead of an extracted script

Extracting the script into `.github/workflows/scripts/` would be the obvious way to make it testable, but a reusable workflow can't reach its own repo's files — `actions/checkout` fetches the *caller's* repo, so the script wouldn't be on disk without a second checkout pinned to a hardcoded ref.

That constraint turned out to be a good thing. The spec loads the workflow, pulls the decision step's `run:` block, and executes it. It tests the bytes that actually ship. A test against an extracted `decide.sh` would keep passing if the workflow started passing it `DENY_LIST` instead of `DENYLIST`; this one fails.

I confirmed the tests have teeth by mutating the workflow three ways — allowing minor updates unconditionally, switching name matching from exact to substring, and renaming an env var. Each mutation failed a different example.

### Gotchas

- The script needs **bash 4** (`readarray`) and **`jq`**. The Buildkite Ruby image is Debian, so bash is fine, but it doesn't ship `jq` — hence the pipeline change.
- macOS ships bash 3.2, so locally the script examples resolve a Homebrew bash, and skip if there isn't one.
- **Skipping is disallowed on CI.** If bash or `jq` are missing there, the examples fail rather than skip, so this file can't quietly stop testing anything.
- The one example that only parses YAML runs everywhere; only the examples tagged `:runs_script` need the toolchain.

### How to test

`bundle exec rake`. 333 examples, 0 failures (19 new).

To see the skip-vs-fail guard: run with `jq` off your `PATH`, once plainly (18 pending) and once with `CI=true` (18 failures).

### DIFF
--- a/spec/workflows/dependabot_auto_merge_spec.rb
+++ b/spec/workflows/dependabot_auto_merge_spec.rb
@@ -1,0 +1,185 @@
+# frozen_string_literal: true
+
+require 'open3'
+require 'tmpdir'
+require 'yaml'
+
+# Drives the decision script exactly as it is embedded in the workflow, rather than a copy of it.
+# A test against an extracted script would keep passing if the workflow stopped passing it the right environment.
+module WorkflowScript
+  WORKFLOW = File.expand_path('../../.github/workflows/reusable-dependabot-auto-merge.yml', __dir__)
+
+  # Bash 4 is the floor because the script uses `readarray`. The Buildkite image is Debian, which satisfies it;
+  # macOS ships bash 3.2, so developers need a Homebrew bash on their PATH.
+  MINIMUM_BASH_VERSION = 4
+
+  module_function
+
+  def workflow
+    @workflow ||= YAML.load_file(WORKFLOW)
+  end
+
+  def step(id)
+    workflow.fetch('jobs').fetch('dependabot-auto-merge').fetch('steps').find { |step| step['id'] == id }
+  end
+
+  def decision_script
+    step('decision').fetch('run')
+  end
+
+  def modern_bash
+    @modern_bash ||= ['bash', '/opt/homebrew/bin/bash', '/usr/local/bin/bash'].find do |candidate|
+      major, = Open3.capture2e(candidate, '-c', 'echo "${BASH_VERSINFO[0]}"')
+      major.to_i >= MINIMUM_BASH_VERSION
+    rescue Errno::ENOENT
+      false
+    end
+  end
+
+  def jq?
+    @jq ||= system('command -v jq > /dev/null 2>&1')
+  end
+
+  # Runs the decision script with the given inputs, returning whether it succeeded and what it decided.
+  def decide(update_type:, dependency_names:, denylist: '[]', minor_allowlist: '[]', merge_method: 'merge')
+    Dir.mktmpdir do |dir|
+      script = File.join(dir, 'decision.sh')
+      github_output = File.join(dir, 'github_output')
+      File.write(script, decision_script)
+      File.write(github_output, '')
+
+      env = {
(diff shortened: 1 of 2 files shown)


spec/workflows/dependabot_auto_merge_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+
+  # Bash 4 is the floor because the script uses `readarray`. The Buildkite image is Debian, which satisfies it;
+  # macOS ships bash 3.2, so developers need a Homebrew bash on their PATH.
+  MINIMUM_BASH_VERSION = 4
+
+  module_function
+
+  def workflow
+    @workflow ||= YAML.load_file(WORKFLOW)
+  end
```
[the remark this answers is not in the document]
iangmaia: I think it sounds reasonable, @mokagio. Wdyt ?

spec/workflows/dependabot_auto_merge_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+        'MINOR_ALLOWLIST' => minor_allowlist,
+        'MERGE_METHOD' => merge_method,
+        'GITHUB_OUTPUT' => github_output
+      }
+      output, status = Open3.capture2e(env, modern_bash, script)
+      step_output = File.read(github_output)
+
+      {
+        success: status.success?,
+        merge: step_output.include?('should-merge=true'),
```
iangmaia, line 64: Right now, anything that exits before writing `should-merge=true` is reported as `merge: false`, so most of the negative examples pass in both cases.

Perhaps `decide` could parse the exact `should-merge` output 🤔  and ordinary denial cases could assert both `success: true` and `merge: false`. Then the malformed-input cases can continue to expect `success: false`.

spec/workflows/dependabot_auto_merge_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+    # Skipping on CI would quietly delete this file's coverage, so fail there instead.
+    raise message if ENV['CI']
+
+    skip message
+  end
+
+  describe 'the decision step' do
+    it 'reads every input the script relies on from the environment' do
+      # Guards the wiring between the workflow and the script: a renamed input must be renamed in both.
+      expect(WorkflowScript.step('decision').fetch('env').keys)
```
iangmaia, line 92: The script examples provide these variables directly, so they don't exercise how the reusable workflow populates them. Maybe it's worth checking the exact expressions assigned to each variable?

Something like:
```ruby
expect(WorkflowScript.step('decision').fetch('env')).to include(
  'UPDATE_TYPE' => '${{ steps.metadata.outputs.update-type }}',
  'DEPENDENCY_NAMES' => '${{ steps.metadata.outputs.dependency-names }}',
  'MINOR_ALLOWLIST' => '${{ inputs.minor-update-allowlist }}',
  'DENYLIST' => '${{ inputs.denylist }}',
  'MERGE_METHOD' => '${{ inputs.merge-method }}'
)
```

mokagio:
@iangmaia thank you for the review. Just a note that I haven't gotten to this yet because of the other "more important" work that's still in play.

PR #138, opened 2026-07-09
Explore Dependabot auto-merge as a Danger plugin

mokagio:
> **Exploration, per @AliSoftware's suggestion on #135** that auto-merge might fit Dangermattic better as a plugin than as a reusable workflow. Draft, not proposed for merge. Same rules as #135, different front-end.

### It works

- **Update type** comes from the YAML document Dependabot appends to its commit message — the same source `dependabot/fetch-metadata` parses. No Action needed. Tested against a real grouped update captured from `woocommerce-android#16232`.
- **Approving** is `github.api.create_pull_request_review`. `github.api` is an Octokit client and `github_utils` already reaches for it.
- **Enabling auto-merge** has no REST endpoint, so it goes through the `enablePullRequestAutoMerge` GraphQL mutation via `Octokit::Client#post('/graphql', …)`.

Where Danger runs matters: every mobile repo triggers Danger on Buildkite with a write-scoped PAT, so the plugin can act. (`reusable-run-danger.yml` forces read-only for Dependabot PRs, but nothing calls it.)

### Tradeoffs, which are the point of this PR

**Danger is a reporter; this makes it an actuator.** Every other plugin observes a PR and leaves a message. This one approves and merges. That's a genuine change in what Dangermattic is, and the strongest argument against.

**Danger runs on events auto-merge doesn't care about** — `labeled`, `milestoned`, `synchronize`. So the plugin needs an idempotency guard, and it has one: it skips approving when it already approved. Enabling auto-merge is idempotent, so that's left unguarded. Not a race so much as noise avoidance, but it is complexity the workflow doesn't carry.

**It only serves repos that run Danger.** Fine for the mobile apps. It does nothing for AutoProxxy, hostmgr, or apps-infra-toolbox, which have no Ruby.

### What I'd conclude

These aren't competing implementations; they're two front-ends onto one rule. The workflow reaches repos with no Ruby. The plugin fits repos that already run Danger and would rather not add a workflow. If both live, the thing worth sharing is the decision logic, not the plumbing — and neither this branch nor #135 does that yet.

I lean toward shipping #135 and treating this as the answer to "could we?" rather than "should we?".

### State

Complete enough to judge, not to merge: 25 examples, `bundle exec rake` green (339 total), `danger plugins lint` passes. **It has never run against a real pull request** — the GitHub calls are all verified doubles, so the shapes are right but the round-trip is unproven. Not documented in the README, and there's no `Dangerfile` example.

I checked the tests bite by mutating the plugin four ways — dropping the already-approved guard, ignoring the allowlist, ignoring the denylist, and dropping the fork check. Each failed at least one example.

### DIFF
--- a/lib/dangermattic/plugins/dependabot_auto_merger.rb
+++ b/lib/dangermattic/plugins/dependabot_auto_merger.rb
@@ -1,0 +1,154 @@
+# frozen_string_literal: true
+
+require 'json'
+require 'yaml'
+
+module Danger
+  # EXPLORATORY. Approves and enables auto-merge on Dependabot pull requests.
+  #
+  # Unlike the rest of Dangermattic, this plugin acts on the pull request rather than reporting on it.
+  # See the pull request that introduced it for the tradeoffs that entails.
+  #
+  # Patch updates are auto-merged. Minor updates are auto-merged only for dependencies on an allowlist,
+  # and a denylisted dependency is never auto-merged, whatever the update type.
+  #
+  # @example Auto-merge patch updates
+  #
+  #          dependabot_auto_merger.auto_merge_patch_updates
+  #
+  # @example Also auto-merge minor updates to a trusted dependency, and never touch a fragile one
+  #
+  #          dependabot_auto_merger.auto_merge_patch_updates(
+  #            minor_update_allowlist: ['release-toolkit'],
+  #            denylist: ['some-fragile-dependency']
+  #          )
+  #
+  # @see Automattic/dangermattic
+  # @tags dependabot, github, process
+  #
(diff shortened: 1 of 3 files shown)


iangmaia:
> Danger is a reporter; this makes it an actuator

> Danger runs on events auto-merge doesn't care about

> It only serves repos that run Danger. 

I think these are strong arguments for not doing it via Danger.
And while plugins potentially running small actions with side effect like adding a missing label would be ok with me (as it's still closely tied with a PR etiquette and validation), I think auto-merging is somewhat an unexpected thing for a plugin to do 🤔

PR #142, opened 2026-08-05
Use Ruby 3.4.9

mokagio:
What is says in the title... See also https://linear.app/a8c/issue/AINFRA-2805


.ruby-version
```diff
@@ -1 +1 @@
-3.2.2
+3.4.9
```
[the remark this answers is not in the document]
mokagio: https://github.com/Automattic/dangermattic/pull/142/commits/b5a10376894ac7c55c46150ffacf3788f88934cc addresses this

image

.buildkite/shared-pipeline-vars
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+RUBY_VERSION="$(cat "$SCRIPT_DIR/../.ruby-version")"
```
AliSoftware: I've seen a lot of repos doing this, and doing this in different ways (eg `$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)` vs just `$(dirname "${BASH_SOURCE[0]}")` vs `$(dirname "$(realpath "${BASH_SOURCE[0]}")")` vs `$(git --show-top-level)` etc). [I actually noticed all those different ways were being used because they broke our `inspect-repos-infra-tool` script](https://github.com/Automattic/apps-infra-toolbox/pull/352) when it was trying to parse those see commands 😅

IMHO this is unnecessary I  the first place, because this script is typically always sourced while the `pwd` is the repo root (since Buildkite runs the build's pipeline steps' commands from the repo root), so we can just reference `.ruby-version` (or `.xcode-version`) bare (without adding a prefix to compute some path relative to the script) and it will work too.

The only time for when trying to compute a relative path would be justified is for the case of monorepos that do some `cd ios` in their step command scripts before building for iOS then later in that script do some manual `source ../shared-pipeline-vars && buildkite-agent pipeline upload conditional-dynamic-steps.yml` themselves in their own scripts. Which is quite rare (I think only beeper does this kind of things calling `source shared-pipeline-vars` themselves from `cd`'d subdirectories today?). But for most repos this dance is unnecessary

And if we want to keep it anyway—even for non-monorepos and repos that never `source` that file manually to upload additional steps dynamically from their own scripts—eg for the sake of consistency across repos and for resilience in case some repos start to source it themselves from subfolders in their future), we should at least standardize the way we do it across repos at some point.

AliSoftware (approved):
Approving but I have disabled auto-merge to leave you the opportunity to address my nitpick on the unnecessary `$SCRIPT_DIR` relative path computation for `.ruby-version`

.ruby-version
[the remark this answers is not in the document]
iangmaia: That's a nice improvement 👍

.buildkite/shared-pipeline-vars
mokagio: Agreed, simplified to a bare `cat .ruby-version` in 9d42640 — Buildkite runs the step commands from the repo root, and this repo never sources the file from a subdirectory.

mokagio: Also created https://linear.app/a8c/project/simplify-version-files-cats-in-buildkite-shared-variables-setup-5da5d7e0efce based on this. Something an AI can work on in the background or during idle time.

PR #144, opened 2026-08-18
Failing specs demonstrating two defects in #112

mokagio:
This PR adds tests to demonstrate possible issues in #112. The tests are obviously failing, as the issues are yet to be addressed. If some or all the issues identified here are worth addressing, the PR can be merged on its base, and the code updated to fix the issues.

1. **A trailing translator comment makes the suggestion delete the string.** `comment_block_containing` (`location_resolver.rb:84`) counts a line that merely *contains* a complete comment as a comment block containing the location, so `start_line == end_line`, `replace_comment` turns on, and the emitted suggestion body is only the new comment. Applying it drops the entry.

2. **An added line starting with `++` shifts every line number after it.** `each_added_diff_line` (`location_resolver.rb:41`) runs its `'+++ '` header filter after the `@@` header is seen, so such a line is skipped without incrementing the counter.

---

### DIFF
--- a/spec/translation_context_checker_spec.rb
+++ b/spec/translation_context_checker_spec.rb
@@ -829,6 +829,56 @@ module Danger
           end
         end
 
+        context 'when the changed entry carries a same-line translator comment' do
+          let(:description) { 'Button that saves the edited settings.' }
+
+          def stub_single_line_entry(path, entry)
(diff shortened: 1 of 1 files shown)


mokagio:
@iangmaia cherry picked the tests into #112. Closing.

PR #145, opened 2026-08-31
Fix cancellation policy and error parsing in the Buildkite retry workflow

mokagio:


mokagio:
## The first cut of defect 1 did not work, and CI has now said so

Validated on the throwaway [#146](https://github.com/Automattic/dangermattic/pull/146), stacked on this branch. Worth recording because the reasoning in the issue — and in my first commit here — was wrong in a way that would have shipped.

Defaulting `cancel-running-github-jobs` to `false` **did not stop the cancellations**. Five pull request events fired about a second apart, five arms, cancelled runs out of five:

| Arm | Group | `cancel-in-progress` | Before | After |
| -- | -- | -- | -- | -- |
| Per-run group | `…-github.run_id` | literal `false` | 0 | 0 |
| Shared group | `…-github.ref` | literal `false` | **3** | **3** |
| Retry workflow, default | `…-github.ref` | `${{ inputs.… }}` | **3 of 6** | **0** |
| Retry workflow, explicit `false` | `…-github.ref` | `${{ inputs.… }}` | **2** | **0** |

The shared-group arm is the one that settles it. It calls nothing, reads no input, and hardcodes `cancel-in-progress: false` — and it still loses 3 of 5 runs. GitHub cancels a previously **pending** run in a shared concurrency group whatever `cancel-in-progress` says, and these runs spend far longer pending than running. So the group was the cause and the flag was never the lever.

That arm keeps cancelling in the "after" column too, which is what stops the zeroes beside it from being an artifact of events that never overlapped. The assert step fails on that case as loudly as on a regression — [it is green here](https://github.com/Automattic/dangermattic/actions/runs/33468956760) reporting `shared group: 3, retry workflow: 0`.

Hence `78c44a7`: the concurrency block is gone rather than keyed per run, which would be a no-op wearing a hat. The input stays declared and ignored, following `label-success-message` on the sibling workflow, because removing it would break every caller that passes it. The spec now pins the absence of the group — **the previous assertion passed against a workflow that still cancelled.**

## Two things this also confirmed on real Actions

- The reusable workflow had **never once been invoked in this repository** (`total_count: 0`) before #146. Nothing in this repo's CI hands the file to GitHub's parser; it is Buildkite only. It now parses, resolves its inputs and runs.
- The API-error path works end to end against a real Buildkite response. With a placeholder token the step logs `❌ Buildkite API call failed: Authentication required. Please supply a valid API Access Token`, so the new `jq -r '.message? // empty'` parsing behaves against a genuine 401 body, not just the stubbed one.

## Still not covered

The empty-build-list and missing-step-key paths need a `200` from Buildkite, and there is no Actions secret in this repo. They stay covered by `spec/github_workflows/`, which runs the identical shell against Debian bash 5.2.37 and jq 1.7.1 — the pairing `ubuntu-latest` provides.

mokagio:
<details>
<summary>Where the cancelled checks actually are — re-measured 2026-09-01, and it is worse than the issue recorded</summary>

Re-ran the rollup across all 14 consumers today. **83 distinct open PRs carrying 198 cancelled `retry-buildkite-job` check runs**, against the 107 recorded on 2026-08-27.

| Repo | PRs hit | Cancelled runs | Rate | Pin | Reached by moving `v1`? |
| -- | --: | --: | --: | -- | -- |
| `pocket-casts-android` | **19** | 42 | 43% | `@v1` | ✅ |
| `pocket-casts-ios` | **15** | **49** | 60% | `@v1` | ✅ |
| `DayOne-Apple` | **12** | 18 | 36% | `@v1` | ✅ |
| `woocommerce-ios` | **11** | **31** | 46% | `@1.2.0` | ❌ |
| `woocommerce-android` | 10 | 20 | 36% | `@v1` | ✅ |
| `simplenote-ios` | 4 | 8 | 67% | `@v1` | ✅ |
| `WordPress-Android` | 3 | 7 | 23% | `@v1` | ✅ |
| `simplenote-android` | 3 | 8 | 47% | `@v1` | ✅ |
| `Gravatar-SDK-Android` | 2 | 6 | **75%** | `@v1` | ✅ |
| `DayOne-Android` | 2 | 4 | 33% | `@v1` | ✅ |
| `Gravatar-SDK-iOS` | 1 | 3 | **75%** | `@v1.1.2.1` | ❌ |
| `WordPress-iOS` | 1 | 2 | 6% | `@v1.1.2` | ❌ |

`release-toolkit` and `a8c-ci-toolkit-buildkite-plugin` currently show none.

### Two corrections to the counts in the issue

**Day One Apple and both Gravatar SDKs are affected and were missing from the original table entirely** — 15 PRs and 27 check runs that were never counted. The WooCommerce repos have also grown sharply since: `woocommerce-ios` 12 → 31 runs, `woocommerce-android` 7 → 20.

One caveat on the headline number: 83 counts distinct PRs and 198 counts check runs, and I cannot tell which of the two the original 107 was. Treat 107 → 198 as indicative rather than exact; the clean before/after is re-running one identical query on both sides.

### Verified example

[`pocket-casts-ios#3550`](https://github.com/Automattic/pocket-casts-ios/pull/3550) carries both outcomes of the same check, which is the shape of the bug — a later run succeeded and the earlier cancelled one stays red forever:

```
☢️ Trigger Danger On Buildkite / dangermattic / retry-buildkite-job: CANCELLED
☢️ Trigger Danger On Buildkite / dangermattic / retry-buildkite-job: SUCCESS
```

Others: [`pocket-casts-android#5211`](https://github.com/Automattic/pocket-casts-android/pull/5211), [`woocommerce-ios#17355`](https://github.com/woocommerce/woocommerce-ios/pull/17355), [`woocommerce-android#15895`](https://github.com/woocommerce/woocommerce-android/pull/15895), [`simplenote-ios#1780`](https://github.com/Automattic/simplenote-ios/pull/1780).

The caller workflow name shows up in the check title — `☢️ Trigger Danger On Buildkite`. That is `github.workflow` resolving to the **caller's** name inside the called workflow, which is why the concurrency group was per consumer repo rather than shared across the org.

### What to watch after the tag moves

**Canaries: `pocket-casts-ios` and `pocket-casts-android`.** Both pin `@v1`, and between them they hold 34 of the 83 affected PRs. If the count does not drop there, the tag move did not take.

**The one pin bump that matters is `woocommerce-ios`.** It is third-worst and the only high-volume consumer the tag move does not reach, because it pins `@1.2.0` — a gem release tag, not a workflow tag. Worth normalising to `@v1` rather than bumping to another immutable pin. `Gravatar-SDK-iOS` and `WordPress-iOS` also need bumps but carry one PR each.

Moving `v1` covers **70 of the 83 PRs and 162 of the 198 runs**. Essentially all of the remainder is `woocommerce-ios`.

</details>

